PR #1, opened 2015-06-17
making the emails microsite aware and changing https to http for bulk…

laq:
… email


felipemontoya:
We would still need  to change "platform_name" for "PLATFORM_NAME" for this to work completely, but we can update it on the db sites to test quickly

felipemontoya:
:+1:

PR #5, opened 2015-08-14
HTML-escape uses of course display name.

laq:
Signed-off-by: Laq <leonardo.quinonez@edunext.co>

Adding the security patch, although it was not applied in the tests that conflicted.


felipemontoya:
We will have to re visit this on the cypress merge, but for now it looks allright :+1:

PR #12, opened 2015-09-18
When emails didnt get the language from the given email they were sen…

laq:
…t in the untranslated form, so now, if there is no language given, it uses the language that is active or current (translations.get_language()), normaly this language cames from the language of the account that launched the email send


laq:
view https://bitbucket.org/edunext/theme/issues/2/mails-in-el-idioma-traducido-ingles

felipemontoya:
:+1:

PR #13, opened 2015-09-18
Supporting the front end of the org_course_creator role

felipemontoya:
This PR modifies the index page of studio so that author can create courses inside the available organizations to them.

- If a user has org specific permissions, the input field is a select instead of a text input


felipemontoya:
@laq could you take a look at this?

PR #17, opened 2015-10-08
Patch to the function that handles the static urls

felipemontoya:
This patch uses startswith('http') . It would fail it the ressource being called would be something like http-image.jpg.
If that happens the browser will try to load a relative path that will fail with a 404 error.


felipemontoya:
@laq

PR #23, opened 2015-10-10
forcing the language as a parameter received in the POST

felipemontoya:


felipemontoya:
Please review and merge @laq

PR #25, opened 2015-10-20
Laq/multilanguage seges

laq:
1. Activating translations in microsites
2. Add danish translation


common/djangoapps/microsite_configuration/backends/database.py
```diff
@@ -99,6 +99,7 @@ def enable_microsites(self, log):
             settings.TEMPLATE_DIRS.append(microsites_root)
             edxmako.paths.add_lookup('main', microsites_root)
             settings.STATICFILES_DIRS.insert(0, microsites_root)
+            settings.LOCALE_PATHS = (microsites_root / 'conf/locale',) + settings.LOCALE_PATHS
```
felipemontoya: If we did this for database defined backends, even if we dont use it, we should do the same for the filesystem backend

laq: Done @felipemontoya

felipemontoya:
+1

PR #26, opened 2015-10-23
Laq/lang pref microsite aware

laq:
Please check what happens here, I thing there can be a more deep discussion about how to make microsite aware the details that keep comming, and to think if the second part of the pull is related to microsite awarenes or simple to relased lang awarenes.


felipemontoya:
For now, lets scope it to just lang awareness and leave the intervention at this point

PR #27, opened 2015-10-27
Arabic Making a pull and a generate from transifex to update the strings

laq:
Updating arabic


laq:
@felipemontoya no it is ready with all languages

felipemontoya:
+1

PR #28, opened 2015-10-29
Fix bug in our ma_lang_user_preference,

laq:
Take in account released_languases is a list of objects
Return none instead of settings.Language_code, so that a later middleware decides the language.


felipemontoya:
:+1:

PR #30, opened 2015-11-10
Merging some of the new translations back into our version

felipemontoya:


felipemontoya:
Adding some of the new translations from upstream arabic into this fork

PR #31, opened 2015-11-10
Adding necessary template tags

felipemontoya:


felipemontoya:
This PR adds two new template tags to support better branding of the wiki pages

PR #37, opened 2015-12-22
Laq/unicode username

laq:
This pull request allows using unicode characters in the username, it is inspired in the commits of edraak for given functionality: 

https://github.com/Edraak/edx-platform/commit/6ab243b64aab0ec53e0bb36f8a0224eae37a35da
https://github.com/Edraak/edx-platform/commit/8e5ca8c097a3fd2fc92ed4b7de0182dcf510e611

This is a simple version with this pitfalls:

* It does not allow spaces: edraak allows them, but in my tests they produce an error with the resolve. We could try to solve it and allow spaces(but i think that's a separate change) or as it was before restrict the spaces.

* Edraak commits also change the admin interface so that it allows changing the unicode usernames through the admin. With this pull request the unicode usernames show fine in the admin, but when changing them it only allows ascii. @felipemontoya  should we include those changes too?

* Edraak also seems to have moved the regex expression to the env file, this could maybe allow the regex to be microsite aware, what do you think?

* Finally: the file edraak_validator could be named edunext_validator, what do you think?

### DIFF
--- a/common/djangoapps/student/edraak_validation.py
+++ b/common/djangoapps/student/edraak_validation.py
@@ -1,36 +1,10 @@
 """
 Allow Unicode in Admin and LMS.
 """
-from django.contrib.auth.admin import UserAdmin
-from django import forms
-from ratelimitbackend import admin
 import re
-from django.contrib.auth.forms import UserCreationForm, UserChangeForm
 from django.core.validators import RegexValidator
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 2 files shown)


felipemontoya:
I say we dont need to admin change them for now.

felipemontoya:
:+1:

laq:
Did you check the final two points? I may have edited the comment after you read it.

PR #38, opened 2016-01-05
Microsite available page

felipemontoya:
When a user lands on a page that is not configured to be a microsite, a new page appears that directs them to the marketing page or in general tells them the domain is not properly configured.

This prevents confusions raised currently by the default microsite.


felipemontoya:
@laq Could you take a look?

laq:
(y)

PR #40, opened 2016-01-21
Fixin the duplicate row on loginFailure table caused by a race condit…

laq:
…ion of the get_or_create on mysql.

catch the error and erase all the rows but one, leaving the one with the newest lockout_until date.


felipemontoya:
Instead of cleaning the entire table.... why not find all objects for that user, add the values, store it, and remove the duplicate?

PR #42, opened 2016-02-19
Dg/ednx/jd 1

jfavellar90:
Making lms xblock runtime microsite aware


felipemontoya:
This PR should be agains the intermediate branch dg/ednx/dogwood+edunext

PR #45, opened 2016-02-25
WIP: redirect middleware

laq:
First setup of the backbone for the redirect middleware, this is missing discussion to make it a full feature.

### DIFF
--- a/common/djangoapps/edunext/middleware.py
+++ b/common/djangoapps/edunext/middleware.py
@@ -5,11 +5,14 @@ import re
 
 from django.conf import settings
 from django.http import HttpResponseRedirect, HttpResponseNotFound
+from django.dispatch import receiver
+from django.db.models.signals import post_save
 
 import edxmako
+from util.cache import cache
+from util.memcache import fasthash
 from microsite_configuration import microsite
 from models import Redirection
-
 
 host_validation_re = re.compile(r"^(?:[0-9]{1,3}\.){3}[0-9]{1,3}(:[0-9]{2,5})?$")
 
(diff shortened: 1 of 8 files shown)


common/djangoapps/edunext/models.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django.db import models
+from django.utils.text import compress_string
+
+from config_models.models import ConfigurationModel
+
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+
+
+class RateLimitConfiguration(ConfigurationModel):
```
felipemontoya: Este nombre que onda?

common/djangoapps/edunext/models.py
```diff
@@ -0,0 +1,23 @@
+"""Models for the util app. """
+import cStringIO
```
felipemontoya: could we purge the imports?

common/djangoapps/edunext/middleware.py
```diff
@@ -0,0 +1,19 @@
+"""
+Middleware for Language Preferences
```
felipemontoya: Eventually we'll get to docstring this better

common/djangoapps/edunext/middleware.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Middleware for Redirecting microsites
+    """
+
+    def process_request(self, request):
+        """
+        Redirect if domain in redirect list
+        """
+        # TODO discuss how and from where this redirect should work, new table? microsites table? sites table?
+        return HttpResponseRedirect('http://www.edunext.co')
```
felipemontoya: just leaving this as an open discussion, should we use `//` to let the protocol be automatic

felipemontoya:
I have updated the feature, tested it on a stage server and added a 5 min cache to the db query

felipemontoya:
@laq any thoughts?

laq:
:+1:  :100:

felipemontoya:
We have config vars for other cache ttls but we never change them. I think i does not make much difference

PR #51, opened 2016-04-04
Fmo/updating da lang

felipemontoya:


felipemontoya:
@juancamilom can you take a quick look and tell me if this looks fine to you?

PR #53, opened 2016-04-18
Moving the execution of the cross-brandig filter to any time after the…

felipemontoya:
… mako middleware has gone through


felipemontoya:
@diegomillan

felipemontoya:
@diegomillan

PR #55, opened 2016-04-25
Supporting redirect of static pages

felipemontoya:
Using a dict to store redirects that can be returned instead of rendering a static template when handling keys in MKTG_URL_LINK_MAP


felipemontoya:
@jfavellar90

PR #57, opened 2016-04-26
Forcing the right path for templates at the template loader

felipemontoya:
Using the microsite.get_template_path on every page load.


felipemontoya:
@Bound3R

Bound3R:
Hi @felipemontoya you should probably squash these commits. :sweat_smile:

PR #60, opened 2016-05-06
This icon is missing for some web-certs

felipemontoya:
the name of the icon is generated using the course mode, and we do generate some certs for audit courses. @Bound3R can you take a look?


Bound3R:
LGTM! :wink:

PR #64, opened 2016-05-25
Jhony/grades discovery

jfavellar90:
First version of enrollments with grades using celery. To get this working, you need to turn on celery for devstack (run command "paver celery --settings=devstack"), and set the value of CELERY_ALWAYS_EAGER to false in your settings file

### DIFF
--- a/common/djangoapps/edunext/routers.py
+++ b/common/djangoapps/edunext/routers.py
@@ -3,10 +3,13 @@
 from rest_framework import routers
 from viewsets import (
     UsersViewSet,
-    CourseEnrrollmentsviewsets,
+    CourseEnrollmentViewset,
+    CourseEnrollmentWithGradesViewset,
 )
 
 
 router = routers.DefaultRouter()
 router.register(r'users', UsersViewSet)
-router.register(r'course-enrollments', CourseEnrrollmentsviewsets)
+router.register(r'course-enrollments', CourseEnrollmentViewset)
+# Async operations
+router.register(r'async/course-enrollments-grades', CourseEnrollmentWithGradesViewset)
--- a/common/djangoapps/edunext/urls.py
+++ b/common/djangoapps/edunext/urls.py
@@ -3,10 +3,13 @@ URLs for the Microsite API
 
 """
 from django.conf.urls import patterns, url, include
+
 from routers import router
+from views import CeleryTasksStatus
 
 
 urlpatterns = patterns(
     '',
     url(r'^data-api/v1/', include(router.urls, namespace='ednx-data-api')),
+    url(r'^data-api/v1/tasks/(?P<task_id>.*)$', CeleryTasksStatus.as_view(), name="celery-data-api-tasks"),
 )
--- a/common/djangoapps/edunext/views.py
+++ b/common/djangoapps/edunext/views.py
@@ -1,0 +1,36 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+from celery.result import AsyncResult
+
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from rest_framework.exceptions import NotFound
+
+from microsite_api.authenticators import MicrositeManagerAuthentication
+
+
+class CeleryTasksStatus(APIView):
+    """
+    view to check celery tasks status
+    """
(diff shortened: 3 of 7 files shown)


common/djangoapps/edunext/routers.py
```diff
@@ -3,10 +3,12 @@
...
-    CourseEnrrollmentsviewsets,
+    CourseEnrrollmentsViewset,
+    CourseEnrrollmentsWithGradesViewset,
 )
 
 
 router = routers.DefaultRouter()
 router.register(r'users', UsersViewSet)
-router.register(r'course-enrollments', CourseEnrrollmentsviewsets)
+router.register(r'course-enrollments', CourseEnrrollmentsViewset)
```
felipemontoya: Im wondering if we should have a completely different endpoint for the async api calls. Also enrollment has 1 r

common/djangoapps/edunext/urls.py
```diff
@@ -3,10 +3,13 @@
...
 from django.conf.urls import patterns, url, include
+
 from routers import router
+from views import CeleryTasksStatus
 
 
 urlpatterns = patterns(
     '',
     url(r'^data-api/v1/', include(router.urls, namespace='ednx-data-api')),
+    url(r'^data-api/tasks/(?P<task_id>.*)$', CeleryTasksStatus.as_view(), name="celery-data-api"),
```
felipemontoya: Should we also have a version number for this?

common/djangoapps/edunext/views.py
```diff
@@ -0,0 +1,36 @@
...
+    def get(self, request, task_id=None, *args, **kwargs):
+        """
+        Return the task status and its result, if already calculated.
+        """
+        if not task_id:
+            raise NotFound()
+
+        task_res = AsyncResult(task_id)
+        # Provisional way to check if a task exists in celery or not
+        # TODO: look for a better way to do this
```
felipemontoya: I found a way to this differnetly at: http://stackoverflow.com/a/10089358/2072496.
IMHO, we gain very little for changing that, so I would leave it like this, but instead of returning the task_id does not exist, just return OK, the task is pending. Is it necessary (at rz or manage) to know that the task id is not know?

common/djangoapps/edunext/viewsets.py
```diff
@@ -57,10 +65,42 @@ class UsersViewSet(DataApiViewSet):
...
+        task_id = "data_api-" + string_now_date + "-" + str(randnum)
+
+        named_args = {
+            "data": serializer.data,
+        }
+        task_instance = EnrollmentsGrades()
+        res = task_instance.apply_async(kwargs=named_args, task_id=task_id)
+
+        data_response = {
+            "task_id": res.id
```
felipemontoya: I would also add the url for checking the task status.

common/djangoapps/edunext/routers.py
jfavellar90: Look at this site https://dev.havenondemand.com/docs/AsynchronousAPI.htm.

common/djangoapps/edunext/views.py
jfavellar90: I was considering the case when Celery service go down for some reason. Rz could stay polling to the the task url with no "done task" answer for long time.

felipemontoya: that is true, but we would also make sure in rz that there is a limit to the time rz keeps polling any data-api endpoint

common/djangoapps/edunext/routers.py
felipemontoya: that is exactly what I was thinking. They have, and we could also, an endpoint of the form `v1/api/sync/course-enrollments`, and one for `v1/api/async/course-enrollments`.

I would start by saying that the default is sync, and in order to highlight that an async operation will take place, we cann add `async/` to the `course-enrollments-grades` endpoint

felipemontoya:
@jfavellar90 Could you please take a look at the type in enrollments, and squash this?

felipemontoya:
That should leave it ready for merge

jfavellar90:
Already rebased and squashed

felipemontoya:
:+1:

PR #66, opened 2016-05-31
Adding the extra xblock video players

felipemontoya:
This PR includes the ooyala and brightclove players to be used as xblocks in our service


felipemontoya:
@laq @jfavellar90

PR #67, opened 2016-06-02
Adding the mentoring requirement explicitly

felipemontoya:
This PR solves a missing dependency for the ooyala player. The xblock originally has it in the requirements.txt file, but not as a install dependency


felipemontoya:
@laq @jfavellar90 @diegomillan

PR #74, opened 2016-06-28
Adding the default setting for this image

felipemontoya:
This PR adds a default setting for the CMS that was present for the LMS already.

This was taken from the [upstream repo] (https://github.com/edx/edx-platform/blob/cfd1620703a6dc567a6aecf73427a84b8161a7ae/cms/envs/common.py#L59)


felipemontoya:
@jfavellar90 can you take a look please?

jfavellar90:
For me it's OK

PR #77, opened 2016-06-30
Fmo/cert fixes ma

felipemontoya:
@jfavellar90 Can you take a post mortem look at this please?

### DIFF
--- a/common/djangoapps/microsite_configuration/middleware.py
+++ b/common/djangoapps/microsite_configuration/middleware.py
@@ -6,13 +6,12 @@ A microsite enables the following features:
 2) Present a landing page with a listing of courses that are specific to the 'brand'
 3) Ability to swap out some branding elements in the website
 """
-import re
-
 from django.conf import settings
 from django.http import Http404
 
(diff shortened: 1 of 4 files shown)


common/djangoapps/microsite_configuration/middleware.py
```diff
@@ -96,31 +95,58 @@ class MicrositeCrossBrandingFilterMiddleware():
...
         # If the course does not belong to an ORG defined in a microsite
         all_orgs = microsite.get_all_orgs()
-        if course_key.org not in all_orgs:
+        if org not in all_orgs:
             return None
 
         # We could log some of the output here for forensic analysis
         raise Http404
+
+    def clean_courses(self, match, request):
```
jfavellar90: You pass the request as a parameter but you don't use it. Same for the function below

felipemontoya: Yes, I did that because it seems plausible that some other filters might need the request object, and I did not want to change the signature of the `clean_xxx` methods later on

PR #79, opened 2016-07-14
Jhony/microsites orgs data migration

jfavellar90:


jfavellar90:
I don't know if the replaced migration will run properly in production, as the system will try to create already existing models

felipemontoya:
@jfavellar90 we would not really run the initial migration, but --fake it instead, and then we do run the second migration

felipemontoya:
:+1: I think this is fine.

PR #82, opened 2016-07-14
simple comment line for show signature

Bound3R:
Hi @felipemontoya, this change is only for show signatures in certificates "honor".


felipemontoya:
:+1:

PR #83, opened 2016-07-14
Changing the permission, to something that can be activated in studio

felipemontoya:
This PR changes the required permissions to access the certificates panel in studio, to the administrator of a course. The administrator (actually `instructor) , can be changed in studio and it is the same permission that we currently use for certificate activation


felipemontoya:
@jfavellar90 FYI

PR #88, opened 2016-07-27
Removing some iexacts

jfavellar90:
This commit is intended to remove iexact lookup types for some fields in report filters. The fields are email, country and mode (course_mode)


jfavellar90:
@juancamilom FYI

juancamilom:
ok :+1:

PR #94, opened 2016-10-25
Fixing grade method call. 

diegomillan:
The signature of this method has changed for eucalyptus release.
You can check new signature [here](https://github.com/eduNEXT/edunext-platform/blob/master/lms/djangoapps/courseware/grades.py#L266)

 @jfavellar90 @felipemontoya


felipemontoya:
Thanks @diegomillan. I think this should be merged asap :+1:

jfavellar90:
:+1:

PR #95, opened 2016-10-26
Org filters for enrollments at the mobile APIs

felipemontoya:
Adding a function that makes it easy to filter enrollments on a microsite level


felipemontoya:
@aliendude Please take a look here and mind this changes

PR #98, opened 2016-10-28
Making edx-notes-api feature microsite aware

jfavellar90:


jfavellar90:
@felipemontoya

felipemontoya:
I assume we need this, right?

jfavellar90:
Yes, it allows to enable or disable the notes service per microsite in lms. For cms it's always activated

felipemontoya:
And the default is enabled or disabled?

jfavellar90:
In lms by default is disabled, in cms by default is enabled

felipemontoya:
:+1:

PR #101, opened 2016-11-04
Adding the xblock_i18n tag to studio

felipemontoya:
Adding the edunext app, so the template tags are available for the django_template renderer.

Solves EDNX-254

@jfavellar90


jfavellar90:
:+1:

PR #105, opened 2016-11-08
add attributes to fix navbar

Bound3R:
This issue has referred in [EDNX-214](https://edunext.atlassian.net/browse/EDNX-214)

### DIFF
--- a/cms/static/sass/elements/_header.scss
+++ b/cms/static/sass/elements/_header.scss
@@ -13,6 +13,7 @@
   header.primary {
     @include box-sizing(border-box);
     @include clearfix();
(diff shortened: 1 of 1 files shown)


cms/static/sass/elements/_header.scss
```diff
@@ -13,6 +13,7 @@
   header.primary {
     @include box-sizing(border-box);
     @include clearfix();
+    @include display(flex);
```
felipemontoya: Are you sure you can do this? does the `display(flex)` is used somewhere else?

felipemontoya:
:+1:

PR #107, opened 2016-11-08
fix: simple add attr value for jquery error

Bound3R:
Is for this reason the button stay blocked... :lock:

### DIFF
--- a/lms/templates/login.html
+++ b/lms/templates/login.html
@@ -100,7 +100,7 @@ from third_party_auth import provider, pipeline
         $submitButton.
           removeClass('is-disabled').
           attr('aria-disabled', false).
-          removeProp('disabled').
+          prop('disabled', false).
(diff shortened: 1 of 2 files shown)


lms/templates/login.html
```diff
@@ -100,7 +100,7 @@
         $submitButton.
           removeClass('is-disabled').
           attr('aria-disabled', false).
-          removeProp('disabled').
+          prop('disabled', false).
```
felipemontoya: Do we need to change this at the register template also? here and in the theme?

felipemontoya: I just saw the theme PR. :+1: 
I'd say we should also fix that here at the `register.html` template

Bound3R: sure, this is for login submit button and the other in theme is of register submit button. :facepunch:

felipemontoya: That is what I mean, lets change it also here, and not only on the theme

PR #109, opened 2016-11-12
Fmo/circle

felipemontoya:
This are the necessary changes to have the CircleCI running in our modified fork.

@juancamilom @jfavellar90 @diegomillan @Bound3R


Bound3R:
@felipemontoya I not tried your code, but after reviewing it briefly I have a question; should we create a test for emails later? 👀

felipemontoya:
@Bound3R, I went trought the commented test again, and it turns out it is a known issue that the test cant run on travis (or circle) because of the html_to_text function. So I flagged it to be skipped in circleci

diegomillan:
It seems good to me :+1:

PR #110, opened 2016-11-15
Bumping the email app version

felipemontoya:
Using the latest email app with the new spanish translations

@jfavellar90 @diegomillan


diegomillan:
:+1: this is the last SHA of email app

PR #113, opened 2016-11-22
Removing the need for the microsite_api_signing_key override. 

felipemontoya:
From now on we will read it from lms.auth.json correctly set by the ansible roles.


felipemontoya:
@jfavellar90

jfavellar90:
:+1:

diegomillan:
:+1:

PR #114, opened 2016-12-01
Reducing the complexity to one indexed query only

felipemontoya:
This PR removes the old logic to use the startswith method on every string, and relies instead on a single indexed query to find the microsite.

@jfavellar90 
@diegomillan 

- [x] Check the configurations in prod to see if there is any site offending relaying on the old logic


diegomillan:
:+1:

jfavellar90:
:+1:

PR #117, opened 2017-01-04
Implementing a cache for the result of the function. 

felipemontoya:
This PR makes it so that in studio, the calls to the microsite.get_value_for_org method use a local cache to improve the performance of loading the /home/ or /assets/COURSE_ID/ pages

The cache lives on the threading.local data so, we need to clear it after each request.

Reviewers
- [x] @jfavellar90 
- [x] @diegomillan


felipemontoya:
I'm working also on a cache for the complete Microsites.objects.all() query

jfavellar90:
It looks good to me

diegomillan:
LGTM :+1:

PR #118, opened 2017-01-09
Fixing the method call. It is an iterable only after being called ()

felipemontoya:
We were passing only the method, not the iterator returned by said method.

For this reason, is that we need to make sure this module has enough test coverage

### Reviewers ###

- [x] @jfavellar90 
- [x] @diegomillan 

Please take a good look, this is something we should have catched on the first review


jfavellar90:
:+1:

jfavellar90:
Was this the reason why the app was failing in your tests?

PR #121, opened 2017-01-19
Jhony/separate microsites app

jfavellar90:
In this PR the idea is to migrate edunext changes to microsite_configuration module to a separated django package. Initially only microsite backends, middlewares and template tags will be migrated.

@felipemontoya 
@diegomillan


felipemontoya:
@jfavellar90 I added a minor change so that the openedx.conf proxy uses the extracted proxy

jfavellar90:
Tests now are OK. This will require a change in our configuration, will do that PR soon

PR #124, opened 2017-02-07
Jhony/microsite model consumers migration

jfavellar90:
What I'm doing here is to remove edunext exclusive modules which have been migrated to edunext-openedx-extensions app. Configuration files were changed in order to get edx-platform working with the new external app. @felipemontoya @diegomillan


jfavellar90:
Circle is kidding me because the bumping of the edunext-openedx-extensions is being ignored, so the tests are run with an old version that doesn't contain the "edunext" subapp

jfavellar90:
The trick was running the tests removing cache. It's an option from circle

PR #125, opened 2017-02-08
edunext-openedx-extensions version bumping

jfavellar90:
As simple as it is: bumping edunext-openedx-extensions version
@felipemontoya 
@diegomillan


felipemontoya:
I think we can merge this, since the tests failed for the known flaky `test_gitlog_pagination_out_of_range_invalid`. Everything else was running.

PR #126, opened 2017-02-10
Bumping version of edunext-openedx-extension to implement the new mic…

jfavellar90:
This PR bumps the version of our external openedx extension app. This version implements the new microsite model.
@felipemontoya 
@diegomillan


jfavellar90:
Complicated to make tests passing, build execution fails because it reaches the maximum memory limit

jfavellar90:
I think this time it failed because a flaky test. Could we ignore that test and approve the PR?

felipemontoya:
I agree, but lets proceed with caution. After merging, the test should pass for master. We will be carefull with deployment
:+1:

PR #133, opened 2017-03-31
Fixing the circle tests

felipemontoya:
This PR uses a flag to output the ga_script. This fixes the third_party_auth test.

@jfavellar90 @Bound3R 

I went with the is_request_in_microsite() method, but we could check for other setting. For example settings.DEBUG, but I was not sure about that. What do yo guys think?

### DIFF
--- a/lms/templates/google_analytics.html
+++ b/lms/templates/google_analytics.html
@@ -2,6 +2,8 @@
 <%!
 from microsite_configuration import microsite
 %>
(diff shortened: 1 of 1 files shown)


felipemontoya:
@Bound3R any comments to this change?

jfavellar90:
:+1: LGTM

Bound3R:
That's ok :wink:

PR #134, opened 2017-04-03
Change in 404 template needed by the smoke tests

jfavellar90:
In this PR I'm adding a hidden input which is read by the smoke tests to identify a page with 404

@felipemontoya 
@Bound3R 

@Bound3R do you agree with this?


felipemontoya:
None of the failing tests are related to this PR, so I approve this PR :+1:

Bound3R:
this input not affecting the platform, approved.

PR #135, opened 2017-04-03
Patching filter_enrollments function for tests failing in production

jfavellar90:
Basically our tests were failing bacause our function filter_enrollments in microsite_aware_functions external app was generating a bug when executing certificates tests. To prevent this to happen, we patched this function for failing tests

@felipemontoya 
@diegomillan


jfavellar90:
You can check the last build of this PR to see that only third party auth test fixed in other PR is failing

felipemontoya:
:+1:

PR #138, opened 2017-04-26
Making the url settings object use the proxy for microsite_configuration

felipemontoya:
This PR makes the urls module of the lms susceptible to changes at the site object.

Merging this will allow us to change the wiki settings on a a per-site basis. This way we can turn it off completely and only activate every site upon request. As a side effect many other urls would also be controllable now.

Reviewers:
@jfavellar90 

FYI:
@juancamilom


jfavellar90:
:+1:, once merged let's deploy and see how it behaves

PR #141, opened 2017-05-10
adding support for scorm

diegomillan:
This PR adds support for scorm through scorm xblock

@jfavellar90 @felipemontoya


felipemontoya:
This is pretty straight forward. :+1:

PR #148, opened 2017-09-04
Delete cookie ecommerce_sessionid from LMS

dfrojas:
This fix the session when a user logout in LMS, closing also the session to in ecommerce.

### DIFF
--- a/common/djangoapps/student/views.py
+++ b/common/djangoapps/student/views.py
@@ -2626,6 +2626,10 @@ class LogoutView(TemplateView):
         else:
             response = redirect(self.target)
 
+        # Get the session cookie of ecommerce and delete it.
+        # Warning: If session cookie in ecommerce settings.py it's changed it,
+        # this won't works.
+        response.delete_cookie('ecommerce_sessionid')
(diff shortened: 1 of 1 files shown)


common/djangoapps/student/views.py
```diff
@@ -2623,14 +2622,18 @@ def dispatch(self, request, *args, **kwargs):  # pylint: disable=missing-docstri
...
         if LogoutViewConfiguration.current().enabled and self.oauth_client_ids:
             response = super(LogoutView, self).dispatch(request, *args, **kwargs)
+
         else:
             response = redirect(self.target)
 
+        # Get the session cookie of ecommerce and delete it.
+        # Warning: If session cookie in ecommerce settings.py it's changed it, 
+        # this won't works.      
+        response.delete_cookie('ecommerce_sessionid')
```
dfrojas: @felipemontoya  I'm deleting the cookie "ecommerce_sessionid" manually and works. But in Open-edx does not  works like this, I did not found other way to fix this bug

felipemontoya:
This PR is not really the best way to solve this, I'm closing now

PR #151, opened 2017-09-06
add specific version of nodeJs

Bound3R:
simply add the nodeJS version as a requirement in the circle ci configuration so that the npm packages were not deprecated.


Bound3R:
the test failed, but it's not due to npm so I'll take the job for granted and merge this pr for no blocking others tests.

PR #158, opened 2017-09-11
[BRT] add templates

Bound3R:
these files add extra features for microsites.
eg `override css`.


Bound3R:
this is a very important change and something big, it requires that we have enabled correctly the feature of microsites...

Bound3R:
@felipemontoya I have several problems here that I haven't been able to solve, could you help me?

PR #162, opened 2017-09-12
gi/ednx/FD-4 Change logo and favicon in LMS and Studio

dfrojas:
This PR change the logos and favicons in Studio and LMS.


felipemontoya:
I thought we had a file for the studio logo with an "open edX STUDIO" version. If not, we probably want a "STUDIO" only img.

Do you know about this @juancamilom

PR #171, opened 2017-09-26
Gi/ednx/FD-3 Allow statics images with http

dfrojas:
This PR allows statics http images

## Reviewers:

- [x] @felipemontoya
- [x] @Bound3R 

## Post-approve action:

- [x] squash

## For testing:

Add a microsite in django admin and in values field add the json object for microsites. In logo_image_url try to pass a http image url. Should show it in the microsite.

### DIFF
--- a/lms/djangoapps/branding/api.py
+++ b/lms/djangoapps/branding/api.py
@@ -321,6 +321,9 @@ def _absolute_url_staticfile(is_secure, name):
         unicode
 
     """
+    if name.startswith("https://") or name.startswith("http://"):
+        return name
+
     url_path = staticfiles_storage.url(name)
 
     # In production, the static files URL will be an absolute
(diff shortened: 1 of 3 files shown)


felipemontoya:
@dfrojas I launch the tests again. It should be a successful build since last time most of the tests passes before the segfault.
@Bound3R can you confirm this is working for you?

lms/djangoapps/branding/api.py
```diff
@@ -321,6 +321,9 @@ def _absolute_url_staticfile(is_secure, name):
         unicode
 
     """
+    if name.startswith("https://") or name.startswith("http://"):
```
Bound3R: I'm not an expert but I think this can be simplified, and it would look something like:
`if name.startswith("http"):`

felipemontoya: but this opens the door to any file which name starts with http. Not cool

felipemontoya: Much as it happens now in static_content

PR #175, opened 2018-01-10
Adding support for gtag in addition to google_analytics

felipemontoya:
# Description

This PR adds support for the gtag script on selected sites. Basically we add another google analytics script if the microsite has the `google_analytics_gtag` value.

# Reviewers
- [x] @Bound3R 
- [ ] @jfavellar90 

# Stakeholders

@Albeiro514

### DIFF
--- a/lms/templates/google_analytics.html
+++ b/lms/templates/google_analytics.html
@@ -22,3 +22,15 @@ from microsite_configuration import microsite
 
 </script>
 % endif
+
+## <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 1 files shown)


lms/templates/google_analytics.html
```diff
@@ -22,3 +22,15 @@
 
 </script>
 % endif
+
+## <!-- Global site tag (gtag.js) - Google Analytics -->
+<% gtag_id = microsite.get_value('google_analytics_gtag', False) %>
+% if gtag_id:
```
Bound3R: I would add a _tabulation_ to this `if`, although it can be omitted anyway...

PR #176, opened 2018-01-26
ADFS metadata has one key for encryption and one for signing

felipemontoya:
# Description
To validate the responses we need the signing key and not only the first key in the metadata

This PR solves exactly that. It is also upstream material, but I want to see it working in production for a while first.

# Reviewers

- [x] @jfavellar90


jfavellar90:
For me it looks good

PR #177, opened 2018-02-16
gi/ednx/LD-9 | microsite aware lms links at studio

andrey-canon:
# Pull Request issue LD-9

## Description
This pull request allows to redirect from Studio to lms microsites. 

## Testing 

- [ ] Check "View live" button which has to set SITE_NAME  value in url.
- [ ] Activate certificates "and checkout Preview" certificate button which has to set SITE_NAME  value in url.
- [ ] Enter to a course in studio site and check out in settings- schedule... that there is a  SITE_NAME  value in url.

## Reviewers:

- [ ] @felipemontoya

### DIFF
--- a/cms/djangoapps/contentstore/utils.py
+++ b/cms/djangoapps/contentstore/utils.py
@@ -5,7 +5,7 @@ Common utility functions useful throughout the contentstore
 import logging
 from datetime import datetime
 
-from django.conf import settings
+from openedx.conf import settings
 from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext as _
 from opaque_keys.edx.keys import CourseKey, UsageKey
@@ -22,6 +22,9 @@ from xmodule.modulestore import ModuleStoreEnum
 from xmodule.modulestore.django import modulestore
 from xmodule.modulestore.exceptions import ItemNotFoundError
 from xmodule.partitions.partitions_service import get_all_partitions_for_course
+
+# eduNEXT 19.11.2015
+from microsite_configuration import microsite
 
 log = logging.getLogger(__name__)
 
@@ -106,15 +109,18 @@ def get_lms_link_for_item(location, preview=False):
 
     # checks LMS_BASE value in site configuration for the given course_org_filter(org)
     # if not found returns settings.LMS_BASE
+    # eduNEXT 19.11.2015 make the link microsite aware, based on the org of the course
     lms_base = SiteConfiguration.get_value_for_org(
         location.org,
         "LMS_BASE",
-        settings.LMS_BASE
+        microsite.get_value_for_org(location.org, 'SITE_NAME', settings.LMS_BASE)
     )
 
     if lms_base is None:
         return None
 
+    # eduNEXT 19.11.2015 Disable preview for now (not microsite aware), it will redirect to a normal course.
+    preview = False
     if preview:
         # checks PREVIEW_LMS_BASE value in site configuration for the given course_org_filter(org)
         # if not found returns settings.FEATURES.get('PREVIEW_LMS_BASE')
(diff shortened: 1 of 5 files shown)


cms/djangoapps/contentstore/utils.py
```diff
@@ -106,15 +109,21 @@ def get_lms_link_for_item(location, preview=False):
 
     # checks LMS_BASE value in site configuration for the given course_org_filter(org)
     # if not found returns settings.LMS_BASE
-    lms_base = SiteConfiguration.get_value_for_org(
```
felipemontoya: This block is most likely equivalent. Inside of the `SiteConfiguration.get_value_for_org` we call `microsite.get_value_for_org`. Could you please test that @andrey-canon. If it is, we can remove the change to make the diff smaller

cms/djangoapps/contentstore/utils.py
```diff
@@ -139,7 +148,14 @@ def get_lms_link_for_certificate_web_view(user_id, course_key, mode):
     assert isinstance(course_key, CourseKey)
 
     # checks LMS_BASE value in SiteConfiguration against course_org_filter if not found returns settings.LMS_BASE
-    lms_base = SiteConfiguration.get_value_for_org(course_key.org, "LMS_BASE", settings.LMS_BASE)
+    lms_base = microsite.get_value_for_org(course_key.org, 'SITE_NAME')
```
felipemontoya: The same for this block. It is most likely already working

common/djangoapps/util/course.py
```diff
@@ -40,14 +43,19 @@ def get_link_for_about_page(course):
     Returns the course sharing url, this can be one of course's social sharing url, marketing url, or
     lms course about url.
     """
-    is_social_sharing_enabled = getattr(settings, 'SOCIAL_SHARING_SETTINGS', {}).get('CUSTOM_COURSE_URLS')
+    is_social_sharing_enabled = getattr(
```
felipemontoya: Please leave this line untouched

common/djangoapps/util/course.py
```diff
@@ -4,7 +4,10 @@
 import logging
 import urllib
 
-from django.conf import settings
+from openedx.conf import settings
+
+from opaque_keys.edx.keys import CourseKey
```
felipemontoya: From what I see, we don't need this, do we?

cms/templates/index.html
```diff
@@ -151,7 +151,15 @@ <h3 class="title">${_("Create a New Library")}</h3>
                 </li>
                 <li class="field text required" id="field-organization">
                   <label for="new-library-org">${_("Organization")}</label>
-                  <input class="new-library-org" id="new-library-org" type="text" name="new-library-org" required placeholder="${_('e.g. UniversityX or OrganizationX')}" aria-describedby="tip-new-library-org tip-error-new-library-org" />
+                  %if org_course_creator_allowed_org:
```
felipemontoya: I know we need this block. But it does not belong to the LD-9 issue. This looks more like BC-14

cms/djangoapps/contentstore/utils.py
felipemontoya: There is one possibility. There is a comment about the microsite module having a cache for the /home/. Both versions should work with the cache eitherway. Let's check and make sure we are not hitting the performance

cms/templates/index.html
```diff
@@ -368,7 +368,7 @@ <h3 class="title">${_("Are you staff on an existing {studio_name} course?").form
           </div>
         </div>
 
-        %if course_creator_status == "granted":
+        %if course_creator_status == "granted" or org_course_creator_status == "granted":
```
felipemontoya: This line also belongs to BC-14

felipemontoya:
Merging since this has passed the tests 4 or so times already. This was just a squash

PR #178, opened 2018-02-19
FEC-0 Support an alternate_platform_name for names 

andrey-canon:
### Description 
This pull request allows a platform name which contains utf-8 characters 

### Test
- Set PLATFORM_NAME with utf-8 values.
- Add ALTERNATE_PLATFORM_NAME without utf-8 values
- Configure your site as a SAML Service Provider
- Check out  the site "your domain"/auth/saml/metadata.xml

### Reviewers 
- [x] @felipemontoya


lms/envs/devstack.py
```diff
@@ -91,7 +91,8 @@
 
 
 def should_show_debug_toolbar(_):
-    return True  # We always want the toolbar on devstack regardless of IP, auth, etc.
```
felipemontoya: This changes are only meant to fix probably pep8 errors, but this forces us to maintain a larger diff.
For this project, quality checks are made only on the lines we are touching, not on the totality of the file. For this reason we try to keep our diff sets to the base minimum

lms/envs/devstack.py
```diff
@@ -114,7 +115,7 @@ def should_show_debug_toolbar(_):
 
 PIPELINE_SASS_ARGUMENTS = '--debug-info'
 
-########################### VERIFIED CERTIFICATES #################################
+########################### VERIFIED CERTIFICATES ########################
```
felipemontoya: the same goes for most changes of this file

lms/envs/devstack.py
```diff
@@ -298,3 +301,5 @@ def should_show_debug_toolbar(_):
 DATA_API_MAX_PAGE_SIZE = 30000
 
 ####################### END DATA API CONFIG #########################
+
+FEATURES["ENABLE_THIRD_PARTY_AUTH"] = True
```
felipemontoya: This change, is something we want to force to true. But it makes more sense to put it before line #234 where also we check for it being true

PR #179, opened 2018-02-19
[ Do not merge ]LD-8 | Unicode Usernames

andrey-canon:
## Description 

Add validation for special characters usernames

## Test 
- Sign up  like a student
- Sign in,edit profile and enroll in a course
- Like staff generate a check out profile information 

# Reviewers
- [ ] @felipemontoya

### DIFF
--- a/common/djangoapps/student/edraak_validation.py
+++ b/common/djangoapps/student/edraak_validation.py
@@ -1,0 +1,15 @@
+"""
+Allow Unicode in Admin and LMS.
+"""
+import re
+from django.core.validators import RegexValidator
+from django.utils.translation import ugettext_lazy as _
+
+
+UNICODE_USERNAME_RE = re.compile(ur'^[\w._+-]+$', re.UNICODE)
+
+VALIDATE_USERNAME = RegexValidator(
+    UNICODE_USERNAME_RE,
+    _("Enter a valid 'username' consisting of letters, numbers, underscores or hyphens."),
+    'invalid'
+)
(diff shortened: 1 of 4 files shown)


felipemontoya:
@andrey-canon could you please remove the changes that are only meant to fix pep8 and pylint, but that have no other use in the context of the PR.

Also, we based this feature on the work done at edraak a long time ago. Since then and precisely from ficus to ginkgo, this [PR](https://github.com/edx/edx-platform/pull/14729) was merged adding support for this. I want to test if we can use that pr instead of this code to get the same functionality.
Perhaps @jfavellar90 wants to be in the loop for this. We could do the changes in stage and decide after we see it in action.

andrey-canon:
@felipemontoya  the issue is fixed in branch ednx/ginkgo+edunext therefore this PR is not necessary,  to test  you only have to add FEATURES['ENABLE_UNICODE_USERNAME'] = True in settings

felipemontoya:
Awesome! Then we need to make sure this feature is turned on, and we can remove completely LD-8 for future migrations :+1:

@jfavellar90 fyi

OmarIthawi:
@andrey-canon @felipemontoya Both `Edraak`'s (being used in [eduNEXT's platform](https://github.com/eduNEXT/edunext-platform)) and the upstream PR (https://github.com/edx/edx-platform/pull/14729) has the same effect. I'd suggest sticking with the upstream new one (https://github.com/edx/edx-platform/pull/14729).

Because there are anticipated changes in Django 1.11 and Django 2.0 which introduces Unicode name support AFAIK.

There's a caveat, you'd have to undo the Edraaks changes first.

felipemontoya:
Hey @OmarIthawi, you are correct, that is what the [Do not merge is all about]. For this migration we have decided to go with the upstream version. Which is good, one less thing to worry about :joy:

felipemontoya:
Yes. This was superseeded by LC-V2 which is now under review as well

PR #187, opened 2018-03-01
Gi/ednx/jd-1

andrey-canon:
## Description 
This branch ensures that SettingsService import from the right dependency.

## Reviewers
- [ ] @felipemontoya


felipemontoya:
In this state, the changes from this PR have been reverted in production. I'm closing this in favor of using JD-1 for the PATCHJD-1 which we still use in prod.

PR #193, opened 2018-04-08
gi/ednx/DFI-5

felipemontoya:
## Description

Allows the value of SEARCH_SKIP_ENROLLMENT_START_DATE_FILTERING to be read from the lms.env.json file

## Commits

Reading edx search configuration to skip or not the enrollment startdate filtering when searching courses

(cherry picked from commit 0450bdc710f2e14e75f8fb6bd4883268dcd3a6f8)


felipemontoya:
I'm merging this w/o tests because the feature is very simple. Tests in master will suffice

PR #196, opened 2018-04-12
Adding settings in lms/aws settings file to guarantee manage and micr…

jfavellar90:
This PR adds the required settings on LMS aws.py configuration file to get microsite and manage API working properly.

@felipemontoya 
@andrey-canon 

Question: In the last migration to ficus, this change was included in [this](https://github.com/eduNEXT/edunext-platform/commit/aff925031604349fe33337284f029444c0a3fa85) commit. It included other changes as well. Now in this migration BC-13 was redefined to contain this change?


felipemontoya:
The answer to your question is: I removed it from PATCHJD-1 because it was not from that scope. We better have this in an issue that we can track, not as a side effect.

felipemontoya:
:+1:

felipemontoya:
@jfavellar90 two options here. If you are confident about this, merge it and deploy to stage.
Second option, deploy this particular branch to stage and test it. Then we come back to merge this.

PR #197, opened 2018-04-13
gi/ednx/GI-11 | Multi-tenant xmodule runtime

felipemontoya:
Using the proxy settings object at the xmodule module_render


felipemontoya:
This hast been also deployed to PROD and it solves the LTI issues

PR #199, opened 2018-04-17
Using the proxy settings object at the xmodule module_render and runtime

felipemontoya:
Backport from commit cfe2f516759ae63eed7cd85a0527d86b0e94ff26

This PR makes sure that the LTI (in fact any xblock) correctly exposes the hostname, so that multi-tenancy is real


felipemontoya:
@juancamilom 
@jfavellar90 
@ericfab179 

This PR worked in stage, I'm merging to solve a support #1966

PR #204, opened 2018-05-04
Sometimes we need to allow enrollments from an environment that does …

felipemontoya:
This PR makes it possible for a site to bypass the org filter for enrollments. This is necessary to allow ecommerce to fullfill orders on the naked site url.


felipemontoya:
Merged w/o tests since the failing ones are flaky. Tests in master will likely pass

PR #205, opened 2018-05-16
Activating the redirection middleware

felipemontoya:
Description
---------------

This PR updates the extensions package and starts using the redirect middelware.

Motivation
-------------
The new version of extensions solves an issue we found on prod with incorrect urls giving 500 errors and has support for the solution of case #2144

Reviewers
-------------

- [x] @jfavellar90 
- [ ] @ericfab179


felipemontoya:
Merged with one approval

PR #206, opened 2018-05-22
JD-1 update

felipemontoya:
# Description
The current configuration will now use None as the seed for non-existant provider_slugs

This means that sites that do not have a `THIRD_PARTY_AUTH_ENABLED_PROVIDERS` will try to find the configuration as they would normally do, but sites that do have the configuration will only use the slugs available at the THIRD_PARTY_AUTH_ENABLED_PROVIDERS dict.

This ultimately fixes the situation where the same provider was available multiple times on the same site.

# Extra info

The JD-1 document in drive holds the remaining info of the feature

# Reviewers
- [x] @diegomillan 
- [ ] @jfavellar90


diegomillan:
LGTM :+1:

PR #207, opened 2018-05-23
Update to JFI-4 | Microsite Middeware Studio

felipemontoya:
# Description
This PR changes studio so that now we use the complete microsite middleware instead of the SimpleMicrositeMiddleware, so we now may have a studio microsite and define settings there.

# Reviewers

- [x] @jfavellar90 
- [x] @diegomillan


jfavellar90:
@felipemontoya so here we can have things like studio.pg.bablabla.com with its own configuration (interesting)? For now we only can support domains like studio.* right?

felipemontoya:
Yes, that is correct. For now, we only want to support regular studio.edunext.co, but have the posibility of adding a specific configuration there.

PR #209, opened 2018-07-27
update recommender

jfavellar90:
Patch to fix a security vulnerability with the recommender xblock

@felipemontoya 
@Alec4r


jfavellar90:
If the tests are OK, to release and deploy

PR #212, opened 2018-10-03
Fmo/support bundle oct 2018

felipemontoya:
This PR bundles a list of minor support cases.

- Adding extra translations for timed exams: 3113
- Making PREVENT_CONCURRENT_LOGINS site aware: 3253
- Adding a lang selector in the footer of studio: 3316
- Bumping version of email-extensions: 3347


common/djangoapps/student/models.py
```diff
@@ -2038,6 +2039,7 @@ def enforce_single_login(sender, request, user, signal, **kwargs):    # pylint:
     Sets the current session id in the user profile,
     to prevent concurrent logins.
     """
+    microsite.set_by_domain(request.META.get('HTTP_HOST', None))
```
jfavellar90: Why do we need to set the microsite here? is there any context where this code is executed and none microsite is set?

jfavellar90:
I only did a comment, for the rest LGTM :+1: @felipemontoya

common/djangoapps/student/models.py
felipemontoya: that is exactly the problem. This never has a microsite already set when running. This happens in paralell where no middleware has been run

PR #214, opened 2018-10-04
Experimental force of the third party auth feature

felipemontoya:
The idea on this PR is to test if we can force a site to only register using third party providers.

Absolutely nothing should happen if the site does not contain the 'ednx_force_third_party_auth' setting


felipemontoya:
@jfavellar90 this is fyi only, if the tests work, we can discuss this in more detail. So far the one thing I really care is that if the feature is off, nothing changes

PR #216, opened 2018-10-04
Updating all the catalogs of js translations

felipemontoya:
We just realized that we had not run the command that actually updates the JS catalogs since moving to ginkgo. The command is `python manage.py lms compilejsi18n` and when running it, we come to this set of changes.

I launched a deploy to staging with this branch, but I would l rather have a bit of a discussion here before merging since the implication is large in the amount of strings, in all the languages and the actual translations


felipemontoya:
Yeah, I left CMS out on purpose, since we don't modify those strings as often and it made the PR twice as largw

felipemontoya:
Got a review from @juancamilom in a different channel. Merging now

PR #219, opened 2018-10-10
Fixing errors at the dasboard if ecommerce fails

felipemontoya:
This PR backports the fix for 500 errors showing at the dashboard when ecommerce was out.

This affects customers that might not have a correct ecommerce url at the time, but whose enrollments might have been paid already.


felipemontoya:
This code was tested on prod, for the problematic case of the customer who reported it and it works.

felipemontoya:
@Ivanca could you please take a look at this today? We would like to merge this soon

felipemontoya:
This are the lines I backported: https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/models.py#L1834

PR #226, opened 2018-11-21
Ha/ednx/bfh - Remove the forum notifier

Squirrel18:
## Description:

This PR moves the conditional block to discussion subscription block, in order to hide the subscription box, only if setting FEATURES['ENABLE_DISCUSSION_HOME_PANEL'] is False.

## Test:

Add to the devstack_docker.py file ENABLE_DISCUSSION_HOME_PANEL = False value in the FEATURES.update() function.

Then the discussion subscription block is not longer shown to the user.

## Reviewers:

 - [ ] @felipemontoya 
 - [ ] @diegomillan

### DIFF
--- a/lms/djangoapps/discussion/static/discussion/templates/discussion-home.underscore
+++ b/lms/djangoapps/discussion/static/discussion/templates/discussion-home.underscore
@@ -6,44 +6,44 @@
     <% } %>
   </section>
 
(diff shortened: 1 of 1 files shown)


Squirrel18:
@felipemontoya @diegomillan Rebased, running test.

PR #229, opened 2018-11-26
ha/ednx/FFI-8_V2

felipemontoya:
This PR is an extension of https://github.com/eduNEXT/edunext-platform/pull/228

Basically after pushing more branches we found that 3 tests from the LMS fail very often in circle while not failing ever on devstack. A better long term solution is required, but right now we need the tests running and failing for the correct reasons (some real feature breaking).


felipemontoya:
I merged this so I could rebase a few other branches waiting in line. Please do take a look and tell me if you know of a better way

PR #236, opened 2018-12-06
ha/ednx/BRT - fix ednx theme 

amalbas:
- Add head-extra.html to load css overrides files. Same content from ginkgo. 

- Change import header.html to navigation.html. To load navigation files from microsites.
old-nav-ha
new-nav ha

### DIFF
--- a/lms/templates/browser-alert.html
+++ b/lms/templates/browser-alert.html
@@ -1,0 +1,14 @@
+<%namespace name='static' file='static_content.html'/>
+<%!
+    from django.utils.translation import ugettext as _
+    from openedx.core.djangolib.markup import HTML, Text
+%>
+
+<!--[if lte IE 9]>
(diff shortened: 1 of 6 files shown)


lms/templates/header.html
```diff
@@ -1,4 +1,4 @@
 ## mako
 <%page expression_filter="h" args="online_help_token"/>
 <%namespace name='static' file='static_content.html'/>
-<%include file="${static.get_template_path(relative_path='header/header.html')}"  args="online_help_token=online_help_token"  />
+<%include file="${static.get_template_path(relative_path='navigation/navigation.html')}"  args="online_help_token=online_help_token"  />
```
felipemontoya: @amalbas are you sure about the implications of this. Does header/header import other files that we need to be aware of in our implementation of navigation/navigation?

Also, would you say that it is time to drop the navigation/navigation and start using header instead for our sites?

amalbas: @felipemontoya there is no problem with the implementation of navigation.html but as it is deprecated I think it's better to start using header.html. It has an improvement in mobile behavior.

amalbas:
@felipemontoya I added missing files from gi/ednx/BRT

PR #237, opened 2018-12-06
ha/ednx/GI-13 | Tighter regex for course filter at index

diegomillan:
This PR makes the regex used for search courses that will land in the homepage more strict. It has to have the complete name and nothing more.

Before this change, an org like EXA would also match in the org EXAMPLE.


felipemontoya:
@diegomillan do you think you can make a OSPR for this? This is a great candidate. We could also add some tests to make it über-clear

diegomillan:
Yes, I do agree on that. I will create an upstream PR @felipemontoya

PR #247, opened 2018-12-19
ha/ednx/BFI-0

amalbas:
Change link and text of copyright studio.
image

### DIFF
--- a/cms/templates/widgets/footer.html
+++ b/cms/templates/widgets/footer.html
@@ -13,7 +13,7 @@ from openedx.core.djangoapps.lang_pref.api import released_languages
 
     <div class="footer-content-primary">
       <div class="colophon">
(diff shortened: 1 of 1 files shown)


cms/templates/widgets/footer.html
```diff
@@ -13,7 +13,7 @@
 
     <div class="footer-content-primary">
       <div class="colophon">
-        <p>&copy; ${datetime.now(pytz.timezone(settings.TIME_ZONE)).year} <a data-rel="edx.org" href="${marketing_link('ROOT')}" rel="external">${settings.PLATFORM_NAME}</a>.</p>
+        <p>&copy; ${datetime.now(pytz.timezone(settings.TIME_ZONE)).year} <a data-rel="edx.org" href="http://www.edunext.co" rel="external">${_("Open edX | Powered by eduNEXT")}</a>.</p>
```
diegomillan: does it make sense to avoid hardcoding the link???  also  we should point to https @amalbas

amalbas: @diegomillan how can we avoid hardcoding the link?

diegomillan: using a setting in studio for example. What do you think @felipemontoya  is it an over killing?

PR #253, opened 2018-12-26
BC-19_PC2 | Updating the edunext plugins

felipemontoya:
This PR updates the eox-* plugins to the latest available


felipemontoya:
Merging now, since tests already passed

PR #256, opened 2018-12-27
BC-19 P6 | Updating xblocks

felipemontoya:
This PR adds xblocks by ibl and proversity


felipemontoya:
merging now since the tests passed already

PR #259, opened 2019-01-08
HA-20 | fix error with the model admin of course enrollment

jfavellar90:
Fix error with the model admin of course enrollment, which was generating 500 error when trying to update a course enrollment record on django admin

@felipemontoya 
@diegomillan

### DIFF
--- a/common/djangoapps/student/admin.py
+++ b/common/djangoapps/student/admin.py
@@ -150,6 +150,7 @@ class CourseEnrollmentForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         super(CourseEnrollmentForm, self).__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


felipemontoya:
I was looking at master, and could not see this change.

https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/admin.py

Is this also failing for master?

jfavellar90:
@felipemontoya I know @Alec4r had a link to the PR fixing this issue, maybe is still pending to merge

felipemontoya:
@Alec4r do you have the link to the original PR?

Alec4r:
@felipemontoya not, i not have the original link of the PR , the pr solved a similar theme but in another model if I remember correctly.

PR #270, opened 2019-01-23
Bumping version of eox-core plugin to 0.12.0

diegomillan:
Bumping version of eox-core plugin to 0.11.0 which includes new business logic to access APIs
@felipemontoya


jfavellar90:
@felipemontoya @diegomillan are we still working on the migration branch ednx/hawthorn+edunext?

diegomillan:
I think so @jfavellar90 @felipemontoya

jfavellar90:
@diegomillan I suppose your changes are not the reason why test went to the recalcada

diegomillan:
Im debugging now , could be a settings problem @jfavellar90

jfavellar90:
@felipemontoya a bump on LLPA of eox-core is required?

felipemontoya:
Test DB continues to fail, but we have tested both locally and in circle over ssh and it passes. I suppose this is related to the cache

PR #276, opened 2019-02-07
Bumping scorm xblock to fix root path issue

diegomillan:
This is PR updates scorm xblock in order to get a root path correctly before uploading content to S3

@felipemontoya 
@jfavellar90 
@Alec4r


jfavellar90:
@diegomillan please make sure the dependency will get installed correctly

diegomillan:
PR updated now. Requirement must be installed in editable mode  @jfavellar90 @Alec4r

PR #277, opened 2019-02-13
allowing to show welcome message using settings

diegomillan:
This PR adds the capability of showing last course update at lms course home, using a custom setting. Also, this change has no side effect on the platform since It would be ignored if the custom setting is not defined.

@felipemontoya 
@Squirrel18 
@Alec4r

### DIFF
--- a/openedx/features/course_experience/views/welcome_message.py
+++ b/openedx/features/course_experience/views/welcome_message.py
@@ -2,6 +2,7 @@
 View logic for handling course welcome messages.
 """
 
+from django.conf import settings
 from django.urls import reverse
 from django.http import HttpResponse, HttpResponseBadRequest
 from django.template.loader import render_to_string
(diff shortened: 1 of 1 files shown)


openedx/features/course_experience/views/welcome_message.py
```diff
@@ -42,7 +43,8 @@ def render_to_fragment(self, request, course_id=None, **kwargs):
             'welcome_message_html': welcome_message_html,
         }
 
-        if get_course_tag(request.user, course_key, PREFERENCE_KEY) == 'False':
+        if (not settings.FEATURES.get('FORCE_SHOW_COURSE_LAST_UPDATE', False) and
```
felipemontoya: I think we should make clear that this FEATURE flag is from us. Normally we do that by calling it EDNX_FORCE....
Also you can put a comment here with the Migration ID that we will use for this. (HA-XX). And create the documentation for the next migration.

On the topic of naming, you could call the Feature EDNX_DISABLE_DISMISS_WELCOME_MESSAGE, which is the name of the view and also the URL.

diegomillan: Feedback applied. I will complete docs also. @felipemontoya

PR #278, opened 2019-02-20
Setting template to show currency defined in the course mode DB record

diegomillan:
This PR modifies the template in order to support custom currencies. Also, we convert a valid ISO 4217 currency code to its corresponding symbol. 

@felipemontoya 
@Squirrel18 
@jfavellar90 
@Alec4r 

Screenshots:
For a course with `usd` as currency

dollar

For a course with `eur` as currency 

eur

### DIFF
--- a/lms/templates/course_modes/choose.html
+++ b/lms/templates/course_modes/choose.html
@@ -104,7 +104,7 @@ from openedx.core.djangolib.markup import HTML, Text
                                                     <ul class="list-actions">
                                                         <li class="action action-select">
                                                             <input type="hidden" name="contribution" value="${min_price}" />
-                                                            <input type="submit" name="verified_mode" value="${_('Pursue a Verified Certificate')} ($${min_price} USD)" />
+                                                            <input type="submit" name="verified_mode" value="${_('Pursue a Verified Certificate')} (${currency_symbol}${min_price} ${currency})" />
                                                         </li>
                                                     </ul>
                                                 </div>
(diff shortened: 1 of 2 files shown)


lms/templates/course_modes/choose.html
```diff
@@ -104,7 +104,7 @@ <h4>${_("Benefits of a Verified Certificate")}</h4>
                                                     <ul class="list-actions">
                                                         <li class="action action-select">
                                                             <input type="hidden" name="contribution" value="${min_price}" />
-                                                            <input type="submit" name="verified_mode" value="${_('Pursue a Verified Certificate')} ($${min_price} USD)" />
+                                                            <input type="submit" name="verified_mode" value="${_('Pursue a Verified Certificate')} (${min_price} ${currency})" />
```
felipemontoya: Is there a way to transform the `currency` variable into a $ or € or <other currency symbol>?

diegomillan: Maybe we could use something like:
http://babel.pocoo.org/en/latest/api/numbers.html#babel.numbers.get_currency_symbol

We still need to pass the correct LOCALE to obtain a corresponding symbol

felipemontoya: That looks good. Do we have babel already installed and available for this mako template?
If yes, I would say, lets do it.
If not, then don't bother, lets remove the $ and move on.

diegomillan: Yes, we already have it. I will modify this in a minute

diegomillan: PR updated now @felipemontoya. See screenshots above for reference

jfavellar90:
Approved XD

PR #280, opened 2019-02-21
Updating social auth

felipemontoya:
This PR updates social-auth-core and social-auth-app-django to the latest available versions as of today.

This will fix:
- errors on the third party auth sad path
- calling the deprecated google plus API

Keep in mind that this version is newer that what the upstream repo has, but we have tested this on a production grade installation with google, facebook linkedin and saml integrations without errors


felipemontoya:
@jfavellar90 @diegomillan @Alec4r Could you please review this and approve if you think its correct?

PR #282, opened 2019-02-22
Updating the version of eox-tenant

felipemontoya:
Updating the version to the latest available v0.3.0 that fixes the ThirdPartyAuth issues


felipemontoya:
@diegomillan and @jfavellar90 If you approve this I will not wait for tests since we want to fix this in PROD now

PR #283, opened 2019-02-28
Updating wiki for security reasons

felipemontoya:
This PR updates the wiki since there was a security alert about XSS in older versions


felipemontoya:
@jfavellar90 ping

jfavellar90:
Would this be an urgent change to apply on our on-premise Hawthorn installations?

jfavellar90:
@felipemontoya why didn't it get merged and deployed? can we proceed?

PR #287, opened 2019-04-15
Updating xblock poll version to 1.8.4

jfavellar90:
Updating the xblock-poll to support correctly the download of poll and survey results through the button "download results to csv". Already tested in stage

@felipemontoya 
@Alec4r 
@diegomillan 
@Squirrel18


Squirrel18:
Is this required to fix some broken behavior?. Upstream is still using 1.5.1. do we need to upgrade this in the next migrations? @jfavellar90

jfavellar90:
@Squirrel18 the motivation behind this PR is [here](https://3.basecamp.com/3966315/buckets/8050706/todos/1647631403). If this is not included in upstream yet, we can create a PR to get these changes merged ASAP. It's sure we have to support this for Ironwood

Squirrel18:
Ok. @jfavellar90

jfavellar90:
Failing tests are unrelated to the applied change so I proceed with the merge

PR #293, opened 2019-06-25
Cross site security upgrades

felipemontoya:
This PR makes it possible to specify a strict policy of login only when the signup source matches the allowed settings


felipemontoya:
This has been tested on STAGE

PR #296, opened 2019-06-27
 Ir/ednx/ha 2

morenol:


lms/envs/ednx_prod.py
```diff
@@ -0,0 +1,16 @@
...
+
+# We intentionally define lots of variables that aren't used, and
+# want to import all variables from base settings files
+# pylint: disable=wildcard-import, unused-wildcard-import
+
+from .production import *
+
+
+# Reading setting to skip or not the enrollment start date filtering for edx-search
+SEARCH_SKIP_ENROLLMENT_START_DATE_FILTERING = ENV_TOKENS.get(
```
morenol: I think we should add this to upstream and remove from this commit @jfavellar90

felipemontoya: This is actually not very useful now, since we can alter this settings during runtime using eox-tenant

morenol:
Since we are using now the plugins to override the settings as we want, does it make sense to have the ednx_prod.py files? @felipemontoya @jfavellar90

cms/envs/ednx_devstack.py
```diff
@@ -0,0 +1,7 @@
+""" Overrides for Docker-based eduNEXT devstack. """
+
+# We intentionally define lots of variables that aren't used, and
+# want to import all variables from base settings files
+# pylint: disable=wildcard-import, unused-wildcard-import
+
+from .devstack_docker import *
```
jfavellar90: Do we run our devstack using these settings??

lms/envs/ednx_devstack.py
```diff
@@ -0,0 +1,17 @@
+""" Overrides for Docker-based eduNEXT devstack. """
```
jfavellar90: Do we run our devstack using these settings??

jfavellar90:
@morenol in fact I don't know why we have those ednx_prod.py files. Are we running from these settings in production? and yeah, with plugin we can override what we want. cc @felipemontoya

cms/envs/ednx_devstack.py
morenol: Yes, but in the case of CMS it is not strictly necessary

lms/envs/ednx_devstack.py
morenol: Yes, we do. We need this for the integration with the other edunext services

jfavellar90:
@morenol it looks good to me

morenol:
@jfavellar90 I am going to squash and wait for the tests to merge this.

PR #297, opened 2019-07-11
bumping eox-tenant version

diegomillan:
This PR bumps eox-tenant with the latest version

@felipemontoya 
@Squirrel18 
@jfavellar90 
@anmrdz


felipemontoya:
We are not upgrading this in the running version yet. I'll leave this open until we decide if we want to get this in afterwards but before the migration to ironwood

diegomillan:
ok

PR #303, opened 2019-07-23
Making student account deletion components a bit site aware

morenol:
I also made a PR against upstream https://github.com/edx/edx-platform/pull/21185/files


jfavellar90:
@morenol the problem we have here is the support of custom translations. We need to link this issue with the work on languages. I suppose we can solve this with Babel, but need to decide if we are going to add all the translations in the languages we support. Maybe it could be an additional feature

jfavellar90:
@morenol I've created IR-2-LANG to remember we have a pending issue here which is translations. I approve this for now.

PR #305, opened 2019-07-24
Language selection LC-V2

cocococosti:
This feature does a match between the languages that the user accepts -on their browser- and the languages available on this tenant. Ednxsite awareness for language selection. It replaces the regular DarkLang app with a ednxsite aware configuration.

### DIFF
--- a/lms/envs/ednx_devstack.py
+++ b/lms/envs/ednx_devstack.py
@@ -44,3 +44,11 @@ GRADES_DOWNLOAD = {
     'BUCKET': 'openedx-grades',
     'ROOT_PATH': '/edx/var/edxapp/uploads',
 }
+
+############################### LANG ########################
+
+LANGUAGE_CODE = 'en'
+TIME_ZONE = 'America/Bogota'
+FEATURES['SHOW_HEADER_LANGUAGE_SELECTOR'] = True
+FEATURES['SHOW_FOOTER_LANGUAGE_SELECTOR'] = True
+FEATURES['EDNX_SITE_AWARE_LOCALE'] = True
--- a/openedx/core/djangoapps/dark_lang/models.py
+++ b/openedx/core/djangoapps/dark_lang/models.py
@@ -2,7 +2,9 @@
 Models for the dark-launching languages
 """
 from config_models.models import ConfigurationModel
+from django.conf import settings
 from django.db import models
+from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
 
 
 class DarkLangConfig(ConfigurationModel):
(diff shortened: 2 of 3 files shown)


lms/envs/ednx_devstack.py
```diff
@@ -16,7 +16,6 @@
 
 LOGGING['loggers']['tracking']['handlers'] = ['tracking']
 
-
```
morenol: Why this line is being removed?

cocococosti: I don't know. I haven't finished editing this PR give me a moment

openedx/core/djangoapps/dark_lang/models.py
```diff
@@ -23,9 +25,22 @@ def released_languages_list(self):
...
 
         Example: ['it', 'de-at', 'es', 'pt-br']
+
+        eduNEXT: we support only the list of available languages from the site
+        otherwise is the same as having no configuration
         """
-        if not self.released_languages.strip():
-            return []
+        if settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False):
+            site_released_langs = configuration_helpers.get_value("released_languages", None)
```
cocococosti: Using an empty list as the default value for released_languages makes the get_value method to always return a dictionary. I solved this in our code as you see here but I think this is a bug from the platform. If there are other parts of our code where empty lists or empty strings are used as default with the get_value method, it could cause errors.

I modified the code from the platform this way: https://github.com/eduNEXT/edunext-platform/commit/74c6190ce62128f0cb9f431d3dba92a461a44974

I tested both options in dev with the LC-V2 feature and it works correctly, but I'm not sure which one is best.

lms/envs/ednx_devstack.py
cocococosti: this line was removed in the original commit I think.

cocococosti:
From what I see from the failed test you should always expect a dictionary from the configuration helpers. So rather than a bug from the platform, I think we should adjust our code...

lms/envs/ednx_devstack.py
jfavellar90: just make sure there's an empty line at the end of the file @cocococosti

openedx/core/djangoapps/dark_lang/models.py
jfavellar90: @cocococosti is this happening from Ironwood? or is it something we bring from hawthorn?

lms/envs/ednx_devstack.py
cocococosti: yes, there's an empty line at the end

openedx/core/djangoapps/dark_lang/models.py
morenol: Why don't you use `list(configuration_helpers.get_value("released_languages", []))`?

lms/envs/ednx_devstack.py
morenol: I think you should add again this line and squash the commits

openedx/core/djangoapps/dark_lang/models.py
morenol: In hawthorn we were using microsite_configuration.microsite instead of configuration_helpers. The problem is on the get_value function on configuration helpers.

jfavellar90: @morenol we can't cast to list because what it returns is a comma separated string of released languages

jfavellar90:
@cocococosti I approved by mistake but please squash the changes

openedx/core/djangoapps/dark_lang/models.py
cocococosti: Because "released_languages" is actually a string, that comment at the docstring is wrong (I'll fix it). What we get from configuration_helpers is a string (or None) that is later parsed into an array.

cocococosti: the comment is not completely wrong but somewhat confusing hahaha I'll leave it as is though

PR #306, opened 2019-07-24
Change logo and favicon in LMS and Studio FD-4

cocococosti:


jfavellar90:
@amalbas what do you think about @morenol comment? handling these logos and favicons from eox-theming?

morenol:
Regarding the favicon, I think that we need to do the change since in the nginx site of the lms (in ironwood) we have this line

  rewrite ^(.*)/favicon.ico$ /static/images/favicon.ico last;

which is preventing those requests from being handled by the lms service

PR #307, opened 2019-07-25
Adding a lang selector in the footer of studio

morenol:


morenol:
Since this is a change on the html template, maybe this could be handled through the theme or anything else we will use for the theming stuffs

jfavellar90:
@morenol I think we can keep this but leaving a note on the migration Excel file about the possible migration to eox-theming

morenol:
I mean, this change can be done without eox-theming. We just need to add the template to the comprehensive theme

morenol:
@jfavellar90 @felipemontoya How we will proceed?

PR #308, opened 2019-07-31
Add support for metadata files that have more than one key.

anmrdz:
### **Description**

This PR adds support for metadata files that have more than one key, this is typical of instances of the ADSF from Microsoft. It modifies the fetching process of the SAML metadata by storing the signing key if there is one specified explicitly.

### **Reviewers**

- [x] @jfavellar90 

- [ ] @morenol 

- [x] @cocococosti


jfavellar90:
@anmrdz since you tested it in dev I approve it

PR #309, opened 2019-07-31
Fix apt key problem

morenol:
With this PR we will prevent problems related to expired apt keys


jfavellar90:
@morenol Ok if that works

PR #310, opened 2019-08-01
Making the regex that gets the index courses more secure GI-13

cocococosti:


jfavellar90:
@cocococosti did you test this one in dev?

cocococosti:
I never knew how. I checked the platform was working correctly but couldn't replicate the error that needed this solution. Big fail on my part, sorry. I'm on it now. @jfavellar90

cocococosti:
Also, is there a unit test for this? I looked for it but couldn't find any.

jfavellar90:
@cocococosti please read the description of the feature and let me know if it's clear for you, if not I can explain you what is this about

cocococosti:
Tested in dev and works correctly. Created two microsites, each with different organizations starting with the same letters: EXA and EXAMPLE. Created a course for each organization. Before the change both courses would show up at the EXA microsite, after the change each microsite shows its own course only. @jfavellar90

jfavellar90:
@cocococosti cool, it's OK from my side

PR #313, opened 2019-08-12
Adding pt_PT

morenol:

PR #314, opened 2019-08-12
Fix tests on circle ci

morenol:
This fixes the tests on the master branch


jfavellar90:
@morenol in the https://github.com/eduNEXT/edunext-platform/pull/313 PR you have this change as well. I suppose you only need to cherry pick one commit from that PR once you merge this

PR #316, opened 2019-08-13
Get urls from microsite settings LD-9

cocococosti:


morenol:
Since we are using eox-tenant and we can override the settings using the eox-tenant signals, is this change still required?

cocococosti:
this feature also disables the preview button, maybe we could move that to eox-theming? or leave it here?

morenol:
Yes, this disables the preview button, but perhaps we do not want to disable that button anymore, do we? @jfavellar90

cocococosti:
Modifying a test to make it pass is not the right thing to do but in this case I think the configuration helpers make 3 extra queries that made the number of queries limit test fail.

jfavellar90:
@morenol this is required because eox-tenant does not work on cms (we have only one "CMS tenant"). On the other hand, for now we don't support preview feature. We can consider to reactivate it once we have multitenancy on CMS, so I think it's better to keep the change as it is. @cocococosti leave the frontend change as it is

jfavellar90:
@cocococosti is it already tested?

cocococosti:
yes, I tested it locally @jfavellar90

PR #318, opened 2019-08-15
Adding MicrositeMiddleware for studio app JFI-4

cocococosti:


jfavellar90:
@cocococosti please move this to eox-tenant

cocococosti:
I thought eox tenant didn't have support for cms. Just to be sure, I should add this to the lms settings at eox tenant? @jfavellar90

morenol:
@cocococosti I think that you should add this to the cms settings at eox tenant. 

The eox-tenant right now is not capable of adding multitenancy to cms, but it can be used to do the same thing that is being done in this PR

cocococosti:
ok, I'll do that @morenol

PR #319, opened 2019-08-18
Updating xblock poll version to 1.8.4

jfavellar90:
This PR updates the version of the xblock-poll to 1.8.4. This version fixes a problem when downloading the poll results as CSV. It requires to add the xblock-poll to the INSTALLED_APPS in django, For this purpose we use the variable ADDL_INSTALLED_APPS in Ansible.

The idea for upstream is to create PR where we add the xblock-poll to the OPTIONAL_APPS by default (common.py), and add the new version to the requirements files. This upstream pr is still pending.

I was wondering if it's better to overwrite the xblock-poll version through Ansible using the extra-requirements variable (it makes this PR unnecessary). What do you think?

@anmrdz 
@morenol 
@cocococosti


jfavellar90:
I think it's better to handle this by keeping these changes

morenol:
I think it is better to add these changes here and use the ansible variables to update the versions in projects that have changes more frequently like our plugins

PR #321, opened 2019-08-23
Add support to absolute http/https urls.

morenol:
This PR adds support to logo routes that starts with http or https in the url.

**TEST**

- In a site configuration/microsite add these settings:

```
"FOOTER_ORGANIZATION_IMAGE": "<Path_to_http/https_logo_url>",
"logo_image_url": "<Path_to_http/https_logo_url>",

```
- Check the logo changes on the tenant.

### DIFF
--- a/common/djangoapps/pipeline_mako/templates/static_content.html
+++ b/common/djangoapps/pipeline_mako/templates/static_content.html
@@ -32,6 +32,8 @@ logger = logging.getLogger(__name__)
 
 <%def name='url(file, raw=False)'><%
 try:
+    if file.startswith("http"):
+        return file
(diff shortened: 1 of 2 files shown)


jfavellar90:
@morenol @cocococosti  I don't know if this is too much for a small change but what about having a custom STATICFILES_STORAGE? This one should only override the url method to validate if the resource name is a url. I think it should go in eox-tenant

morenol:
@jfavellar90 I think that it is a reasonable idea. In that case, I think that for this feature,  eox-core is more appropriate.

jfavellar90:
@morenol agree, going to eox-core

PR #324, opened 2019-09-09
Add Edunext requirements to Circle CI.

cocococosti:
Installing edunext requirements at the circle ci env to fix import errors when running tests in FFI-3.

Is the name of the feature correct? FFI-8? or should this be part of BC-19? @morenol @jfavellar90

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
 
       - restore_cache:
           keys:
-            - v1-{{ checksum ".circleci/config.yml" }}-pip-deps-{{ checksum "requirements/edx/base.txt" }}-{{ checksum "requirements/edx/testing.txt" }}-{{ checksum "requirements/edx/django.txt" }}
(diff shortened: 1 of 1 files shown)


morenol:
Did you run the tests of FFI-3 with this change? Are they ok?

jfavellar90:
@cocococosti I agree, this is part of FFI-8. The version of eox-tenant in the file requirements/edunext/base.txt is an outdated one, but I think I should work. I suppose you need to merge this and then rebase the FFI-3

cocococosti:
Yes, I need to merge this first and then rebase with FFI-3.

morenol:
I am not sure if this if FFI-8, since this depends on BC-19 and BC-19 depends on FFI-8. I think this should have its own name.

I think that we should not merge this, until we are not sure that FFI-3 tests are OK.

We should rebase FFI-3  with this branch  to run the tests

cocococosti:
Yes, you're right I also think it should have a name of its own since it depends on both BC 19 and FFI 8. Maybe leave it as part of FFI-3? I think is the only feature that needs the ednx requirements installed in circle ci.

I'll need to create a new PR to change the branch name. 

What do you think? @jfavellar90

jfavellar90:
@cocococosti @morenol  I hope this is going to disappear on the next release. You're right on the fact that we must handle this one as a separated feature, depending on FFI-8 and BC-19. Proceed with the branch name change

PR #325, opened 2019-09-09
Removing annoying confirmation message

felipemontoya:
This PR applies the patch from:

https://github.com/edx/edx-platform/commit/881c8e035c57984ab3ddab2b816613f13c1eb539

And also removes `edx` from the message string.

This solves the 6873 support ticket.


felipemontoya:
@jfavellar90 this is a candidate for **things we need to migrate to ironwood**

jfavellar90:
@felipemontoya added on a migration TODO

PR #327, opened 2019-09-10
ir/ednx/IR-4

cocococosti:
Applies a patch (this part should disappear in the next release right) and removes the 'edx' word from the message.


morenol:
Will we create a PR that removes the 'edx' string to upstream?

cocococosti:
Maybe we should do that.
Could we put a variable there that takes the name of the platform from the settings instead of Edx?

cocococosti:
ok to merge this? @jfavellar90

PR #330, opened 2019-09-23
Fix to prevent 500 error when course mode currency is not the same as…

morenol:
The problem arises when the tenant has defined PAID_COURSE_REGISTRATION_CURRENCY with a different currency than the currency defined in the course mode.


felipemontoya:
Is this something that already exists on master? or are we adding this as a modification that we need to keep track of?

morenol:
This does not exists on our master. We applied this in other projects https://github.com/eduNEXT/edx-platform/commit/522d1941ea0e1f804da01d85f107dc197ae98a12

jfavellar90:
This was a quick fix I applied to get rid of this error. I approve this PR, but then we'll have to discuss about the better way to handle multi-currency in LMS/ecommerce

PR #331, opened 2019-09-27
Activation emails handled by ACE

jfavellar90:
This PR enables HMTL format for activation emails. This is a back-port of [this upstream PR](https://github.com/edx/edx-platform/pull/18928). In fact that upstream PR was reverted later because an issue with the serialization of the message object in the celery task in charge of sending the activation email. This PR takes care of that by running the sending synchronously.

**NOTE**
After merge, a languages compilation is required to localize strings in added django templates

@felipemontoya 
@cocococosti 
@morenol


jfavellar90:
@felipemontoya in fact no, the translations need to be added. In hawthorn prod we have this problem because multi language activation email are not supported. I just asked Albeiro about related support cases, his answer is still pending

jfavellar90:
@felipemontoya @morenol confirmed, this is an open issue in Hawthorn

PR #335, opened 2019-10-07
Applying patch to enable password change from django admin

cocococosti:
This PR applies a patch (from https://github.com/edx/edx-platform/pull/21616) to enable user password change from the django admin page with a feature setting.

Tested in dev and works OK.

### DIFF
--- a/cms/urls.py
+++ b/cms/urls.py
@@ -208,8 +208,9 @@ if settings.FEATURES.get('AUTH_USE_CAS'):
     ]
 # The password pages in the admin tool are disabled so that all password
 # changes go through our user portal and follow complexity requirements.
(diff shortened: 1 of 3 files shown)


jfavellar90:
LGTM @cocococosti

cocococosti:
Setting ENABLE_CHANGE_USER_PASSWORD_ADMIN directly in the microsite settings didn't work locally, it was taking a None value.

@morenol says it may be because the setting is initialized before the eox tenant signal.

It worked when I added the setting directly in lmv.env.json with a true value so it'll work if we add the setting directly in customer data, fyi @jfavellar90

jfavellar90:
@cocococosti this is not a tenant-aware feature flag. We activate it globally, reason why we must manage this via ansible

cocococosti:
ok, then I'll merge once the tests pass

PR #336, opened 2019-10-07
Bumping eox-tenant version to 1.2.9

jfavellar90:
Bumping eox-tenant version to 1.2.9

@cocococosti 
@morenol


jfavellar90:
This is not gonna break the tests at all (in the new eox-tenant version just added a new config in aws.py file)

PR #337, opened 2019-10-08
Using social_core in version 3.2.0

jfavellar90:
Using social_core in version 3.2.0 to fix problem with linkedin third party auth

@felipemontoya 
@morenol 
@cocococosti


morenol:
Did this work on stage?

jfavellar90:
@morenol we go directly to prod with this change

PR #346, opened 2019-11-01
Applying patch plus extra fixes to unicode strings

cocococosti:
Applying patch from https://github.com/edx/edx-platform/commit/6e81c84d8bdef9b1eb70683a805ea20e6120b854 plus extra fixes to cohort strings:

When the cohort name has a unicode character, for example "promoción", the "Show email sent history" functionality at the instructor dashboard (lms) gives the following error:

` exceptions:UnicodeEncodeError: 'ascii' codec can't encode character u'\xf3' in position 28: ordinal not in range(128) `

The email history contains strings with the cohort name that were not unicode scaped. 

To solve this I scaped all strings with the cohort name, the others strings had the course ID and email which won't contain any special character.

Selection_006

Tested in dev and works ok.

This solves ticket 7850.

### DIFF
--- a/lms/djangoapps/bulk_email/models.py
+++ b/lms/djangoapps/bulk_email/models.py
@@ -152,10 +152,10 @@ class CohortTarget(Target):
         return self.short_display()
 
     def short_display(self):
-        return "{}-{}".format(self.target_type, self.cohort.name)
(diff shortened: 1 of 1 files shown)


cocococosti:
Couldn't apply the patch, the test were failing because of previous changes made to the platform in upstream. 
Just leaving the changes necessary to solve the ticket, this should disappear with the next migration.

PR #354, opened 2019-11-21
Do not hide all the elements of the units with problems to anonymous …

morenol:
With this change, the units with problem components are not hidden

image

### DIFF
--- a/common/lib/xmodule/xmodule/seq_module.py
+++ b/common/lib/xmodule/xmodule/seq_module.py
@@ -465,13 +465,6 @@ class SequenceModule(SequenceFields, ProctoringFields, XModule):
             item_type = item.get_icon_class()
             usage_id = item.scope_ids.usage_id
 
-            if item_type == 'problem' and not is_user_authenticated:
-                log.info(
-                    'Problem [%s] was not rendered because anonymous access is not allowed for graded content',
-                    usage_id
-                )
-                continue
-
             show_bookmark_button = False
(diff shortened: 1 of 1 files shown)


common/lib/xmodule/xmodule/seq_module.py
```diff
@@ -465,13 +465,6 @@ def _render_student_view_for_items(self, context, display_items, fragment, view=
             item_type = item.get_icon_class()
             usage_id = item.scope_ids.usage_id
 
-            if item_type == 'problem' and not is_user_authenticated:
```
cocococosti: Just a question: the whole block is classified as "problem" if there is a graded element in it?

morenol: It is being used the icon class to define the type of unit. If there is a component in the unit with 'problem' as icon class then the icon class of the unit is "problem". Those cases are the ones that are failing.

cocococosti: ah yes, they said it in the comments, getting the type from the icon class is kinda weird

felipemontoya:
I suggest that we wait for the upstream PR to be merged and then do a backport before we continue with this. The idea is that we don't deviate too much from upstream here so that we can support this better in the future.

morenol:
**Update**: this is the PR against upstream: https://github.com/edx/edx-platform/pull/22385

PR #357, opened 2019-12-03
Revert "Add translations from sga xblock"

morenol:
Reverts eduNEXT/edunext-platform#356


morenol:
This is not the problem. The actual issue is described in  https://github.com/mitodl/edx-sga/issues/252

PR #358, opened 2019-12-04
Update edx-sga

morenol:
Update edx-sga version, which is basically the same version that we already had + this change https://github.com/eduNEXT/edx-sga/commit/97d41c3fcc3c641a9ed5262b9f4cec0589a06b50

This change allows us to copy units with the edx-sga xblock.

This change is already in the upstream version but I am not installing that version because that version is not 100% compatible with ironwood


morenol:
Merging since it is a problem in prod

felipemontoya:
I only now see this, but I agree that the merging needed to happen

PR #381, opened 2020-03-26
Redirect to a custom page after user activation

morenol:
When a user activates her account, she is redirected to /dashboard.

This change allows us to make a custom redirection.


morenol:
@jfavellar90 @Alec4r  Can you take a look to this PR?

PR #388, opened 2020-04-08
Upgrade eox-core

morenol:


morenol:
Merging to fix error from the previous release.

PR #391, opened 2020-04-14
Workaround for SAAS-44 until we migrate to python3.

morenol:


felipemontoya:
Do we need to do anything about the way that the session data is written to the cookie? to make sure we don't make more `\"` cookies?

felipemontoya:
There are a few extra lines that can be removed. And we have not yet addressed the write error. Why are there "" in the first place? Do you get those in the response using devstack?

PR #392, opened 2020-04-15
Upgrade eox core

morenol:


felipemontoya:
I have a change on eox-core that I need to upload to prod today.
Please lets wait for an hour and I'll get it ready on a new eox-core release

felipemontoya:
I think the error has absolutely no relation:

```
    def test_enrollment_track_partition_added(self):
        """
        Test that the dynamic enrollment track scheme is added if there is no conflict with the user partition ID.
        """
        all_partitions = get_all_partitions_for_course(self.course)
>       self.assertEqual(2, len(all_partitions))
E       AssertionError: 2 != 1
```

@morenol are you ok if we merge like this?

felipemontoya:
Better yet, you re-ran them.

PR #393, opened 2020-04-17
Install tpa dependecies of eox-core

morenol:
Related to https://github.com/eduNEXT/eox-core/pull/92/files

We still need to create that tag on eox-core. This will install the eox-core dependecies for tpa


morenol:
I saw that in the tests python-jose is not being installed, I think that it is happening because the [version](https://github.com/python-social-auth/social-core/blob/1.7.0/requirements-openidconnect.txt) used in tests doesnt have that version as requirement.

morenol:
Not, it install python jose in the tests https://circleci.com/gh/eduNEXT/edunext-platform/6238

PR #398, opened 2020-05-07
Bumping eox-core version to v2.12.3

jfavellar90:


requirements/edunext/base.txt
```diff
@@ -58,4 +58,4 @@ git+https://github.com/proversity-org/badgr-xblock.git@7077d499b2f752c78bb41e9c1
 ###################
 # plugins must be editable for the settings to appear on the sys path during loading
 -e git+https://github.com/eduNEXT/eox-tenant.git@v2.5.1#egg=eox-tenant==2.5.1
-git+https://github.com/eduNEXT/eox-core.git@v2.12.2#egg=eox-core[sentry,tpa]==2.12.2
```
morenol, line 61: I think that the version in eox-core should be 2.13.0 because it seems to me that it is not backward compatible, but anyway that is not related to this PR

jfavellar90: @morenol I decided to make it backwards compatible: https://github.com/eduNEXT/eox-core/commit/1089791fc906f9fe786aac9cf376c864794423d5

PR #403, opened 2020-07-17
Making MAX_ASSET_UPLOAD_FILE_SIZE_IN_MB configurable

felipemontoya:
Adding the capacity to modify this limit via Ansible


morenol:
Shouldn't we add this in eox-core?

felipemontoya:
Why eox-core? ideally we should not need to do a PR like this to override a global settings value. I'm all for making eox-core support all the ENV.keys, but that would be a more complex PR

PR #405, opened 2020-07-31
Pre enrollment hook

felipemontoya:
This is the first implementation of the concept described at:
https://docs.google.com/document/d/1jhnudz6AVtVt0ZSRSwOwj9gJ0lsDDn_8mUCPehLPzLw/edit

This PR only adds the pre_enrollment trigger.

### DIFF
--- a/common/djangoapps/student/models.py
+++ b/common/djangoapps/student/models.py
@@ -1321,11 +1321,11 @@ class CourseEnrollment(models.Model):
         except Exception as error:  # pylint: disable=broad-except
             raise TriggerException(error.message)
 
-        for receiver, response in hook_result:
+        for receiver_func, func_response in hook_result:
             log.info(
                 u"This is the result of calling the pre_enrollment signal. Receiver was %s and response %s",
-                receiver,
-                response,
+                receiver_func,
+                func_response,
             )
 
         if mode is None:
--- a/common/djangoapps/student/models.py
+++ b/common/djangoapps/student/models.py
@@ -66,6 +66,7 @@ from openedx.core.djangoapps.content.course_overviews.models import CourseOvervi
 from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
 from openedx.core.djangoapps.xmodule_django.models import NoneToEmptyManager
 from openedx.core.djangolib.model_mixins import DeletableByUserValue
+from openedx.core.lib.triggers.v1 import pre_enrollment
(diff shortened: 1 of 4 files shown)


common/djangoapps/student/models.py
```diff
@@ -1310,7 +1311,20 @@ def enroll(cls, user, course_key, mode=None, check_access=False):
         verified the user authentication.
 
         Also emits relevant events for analytics purposes.
+
+        This method triggers a public signal that may prevent the creation of the enrollment.
         """
+
+        # This signal allows plugins to extend the behaviour of the enrollment process.
+        hook_result = pre_enrollment.send(sender=None, user=user, course_key=course_key, mode=mode)
```
mariajgrimaldi, line 1319: how would look like the `result` using the current receiver in eox-hooks?

felipemontoya: If you want to stop the enrollment from happening, you need to raise an exception of any kind

common/djangoapps/student/views/management.py
```diff
@@ -384,8 +384,11 @@ def change_enrollment(request, check_access=True):
                 enroll_mode = CourseMode.auto_enroll_mode(course_id, available_modes)
                 if enroll_mode:
                     CourseEnrollment.enroll(user, course_id, check_access=check_access, mode=enroll_mode)
-            except Exception:  # pylint: disable=broad-except
-                return HttpResponseBadRequest(_("Could not enroll"))
+            except Exception as error:  # pylint: disable=broad-except
+                message = error.message
+                if not message:
+                    message = _("Could not enroll")
+                return HttpResponseBadRequest(message)
```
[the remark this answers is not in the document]
felipemontoya: I changed it so that this only happens on TriggerException (defined in the same file as the signals).

The place where we send the actual signal will be wrapped in a try catch. If anything happens there, this TriggerException will be raised. Anything else remains as it was before

felipemontoya:
@andrey-canon and @mariajgrimaldi I opened ssh sessions to the failing tests suites for lms and cms.
The error message behind every failure is:
VersionConflict: (path.py 11.5.2 (/tmp/workspace/venv/lib/python2.7/site-packages), Requirement.parse('path.py<9.0.0,>=8.2.1'))

Since this is not related to anything we are doing here, Im taking it as a cache problem.

I'll fix the quality issues and if you are ok I'll procede to merge

PR #408, opened 2020-09-02
Adding quick fix to solve the SGA issue with grading

cocococosti:
This PR is to quickly fix the sga issue reported here https://discuss.openedx.org/t/problems-with-sga-grade-submission/2415

A better solution should be applied in the future.

I didn't test it locally but this fix is a copy of the one made by @Squirrel18 in Pearson, where it works correctly.


cocococosti:
yes, it is technically correct.

jfavellar90 (approved):

spokerman12 (approved):

PR #416, opened 2020-09-25
ju/ednx/LC-V2: Ednxsite awareness for language selection

mariajgrimaldi:
**Description**

This feature does a match between the languages that the user accepts -on their browser- and the languages available on this tenant. Ednxsite awareness for language selection. It replaces the regular DarkLang app with a ednxsite aware configuration.

**Tests**

Following the Feature Document, added LANGUAGE_CODE and released_languages to a test microsite:

```
"LANGUAGE_CODE":"en",
"released_languages":"en,ru,fr,es-419,de-DE,pt-BR"
```

**Tests results**

LMS:

Screenshot from 2020-09-25 10-59-39

STUDIO:

Screenshot from 2020-09-25 11-10-17

**Can we remove this feature of the platform?**

-After talking with @andrey-canon  -
 
I thought about moving this feature to eox-tenant or another plugging where it could make sense, it would be something like:

* Inherit from `DarkLangMiddleware` and override process_request as is done in this PR. Then, place this new middleware in the same spot as DarkLangMiddleware
* Make a proxy model using `DarkLangConfig` and override `released_languages_list` as is done in this PR

I wanted to start a discussion about if this change is worth it or not before trying it.

### DIFF
--- a/openedx/core/djangoapps/dark_lang/middleware.py
+++ b/openedx/core/djangoapps/dark_lang/middleware.py
@@ -84,9 +84,10 @@ class DarkLangMiddleware(MiddlewareMixin):
 
     def process_request(self, request):
         """
-        Prevent user from requesting un-released languages except by using the preview-lang query string.
+        eduNEXT: this middleware will always process the requests.
+        The model has been modified to not even attempt the db lookup
         """
-        if not DarkLangConfig.current().enabled:
+        if not settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False) and not DarkLangConfig.current().enabled:
             return
 
         self._clean_accept_headers(request)
@@ -117,8 +118,13 @@ class DarkLangMiddleware(MiddlewareMixin):
         Remove any language that is not either in ``self.released_langs`` or
         a territory of one of those languages.
         """
+        ednx_locale = settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False)
         accept = request.META.get('HTTP_ACCEPT_LANGUAGE', None)
         if accept is None or accept == '*':
+            if ednx_locale:
+                # eduNEXT: return the site aware settings.LANGUAGE_CODE
+                # so that django.utils.locale.LocaleMiddleware can pick it up
+                request.META['HTTP_ACCEPT_LANGUAGE'] = "{};q={}".format(settings.LANGUAGE_CODE, "0.1")
             return
 
         new_accept = []
@@ -127,6 +133,9 @@ class DarkLangMiddleware(MiddlewareMixin):
             if fuzzy_code:
                 # Formats lang and priority into a valid accept header fragment.
                 new_accept.append("{};q={}".format(fuzzy_code, priority))
+            elif ednx_locale:
+                # eduNEXT: if there is no match, we set it to the settings.LANGUAGE_CODE
+                new_accept.append("{};q={}".format(settings.LANGUAGE_CODE, "0.1"))
 
         new_accept = ", ".join(new_accept)
 
--- a/openedx/core/djangoapps/dark_lang/models.py
+++ b/openedx/core/djangoapps/dark_lang/models.py
@@ -4,6 +4,7 @@ Models for the dark-launching languages
 
 
 from config_models.models import ConfigurationModel
+from django.conf import settings
 from django.db import models
 from django.utils.encoding import python_2_unicode_compatible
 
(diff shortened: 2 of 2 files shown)


openedx/core/djangoapps/dark_lang/middleware.py
```diff
@@ -117,8 +119,13 @@ def _clean_accept_headers(self, request):
         Remove any language that is not either in ``self.released_langs`` or
         a territory of one of those languages.
         """
+        ednx_locale = settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False)
```
andrey-canon, line 122: This could be the first line and use ednx_locale in the line  90

openedx/core/djangoapps/dark_lang/middleware.py
```diff
@@ -84,10 +84,12 @@ def beta_langs(self):
...
         """
-        Prevent user from requesting un-released languages except by using the preview-lang query string.
+        eduNEXT: this middleware will always process the requests.
+        The model has been modified to not even attempt the db lookup
         """
-        if not DarkLangConfig.current().enabled:
-            return
+        if not settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False):
+            if not DarkLangConfig.current().enabled:
+                return
```
andrey-canon, line 92:
```suggestion
        if not settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False) and not DarkLangConfig.current().enabled:
            return
```

openedx/core/djangoapps/dark_lang/models.py
```diff
@@ -37,9 +38,21 @@ def released_languages_list(self):
...
+            site_released_langs = getattr(settings, "released_languages", None)
+
+            # If there aren't any released_languages configured
+            if not site_released_langs:
+                site_released_langs = []
+            else:
+                site_released_langs = [lang.lower().strip() for lang in site_released_langs.split(',')]
+                site_released_langs.sort()
+
+            return site_released_langs
```
andrey-canon, line 55:
```suggestion
        released_languages = self.released_languages

        if settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False):
            site_released_langs = getattr(settings, "released_languages", None)            
            released_languages = site_released_langs if site_released_langs else  []
```

and change the line 57 to `languages = [lang.lower().strip() for lang in released_languages.split(',')]`

Thoughts ??

andrey-canon (changes requested):

openedx/core/djangoapps/dark_lang/models.py
mariajgrimaldi: good idea, I'll try it

mariajgrimaldi: I did this:
```
        released_languages = self.released_languages

        if settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False):

            site_released_langs = getattr(settings, "released_languages", None)
            released_languages = site_released_langs if site_released_langs else ""

        languages = [lang.lower().strip() for lang in released_languages.split(',')]

        # Put in alphabetical order
        languages.sort()

        return languages
```
I changed `else []` for `else ""` so the `.split()` won't fail

mariajgrimaldi: For what I tested, the behavior is the same, when `site_released_langs` is `None` this happens in both cases:

Screenshot from 2020-09-30 11-32-15
So I guess the change is OK

openedx/core/djangoapps/dark_lang/middleware.py
mariajgrimaldi: using `self.ednx_locale`?

openedx/core/djangoapps/dark_lang/models.py
```diff
@@ -37,13 +38,22 @@ def released_languages_list(self):
...
+
+        eduNEXT: we support only the list of available languages from the site
+        otherwise is the same as having no configuration
         """
-        if not self.released_languages.strip():
-            return []
+        released_languages = self.released_languages
+
+        if settings.FEATURES.get("EDNX_SITE_AWARE_LOCALE", False):
+
```
andrey-canon, line 48: unnecessary space

openedx/core/djangoapps/dark_lang/models.py
```diff
@@ -37,13 +38,22 @@ def released_languages_list(self):
...
+            site_released_langs = getattr(settings, "released_languages", None)
+            released_languages = site_released_langs if site_released_langs else ""
+
+        languages = [lang.lower().strip() for lang in released_languages.split(',')]
```
andrey-canon, line 52: remove the blank space after this in order to do the diff smaller

openedx/core/djangoapps/dark_lang/middleware.py
andrey-canon: @mariajgrimaldi

andrey-canon (changes requested):

openedx/core/djangoapps/dark_lang/middleware.py
mariajgrimaldi: what do we gain doing that? it will be used just once in process_request

andrey-canon: yes, you are right github doesn't show the whole file and I thought it was  in the same method

PR #417, opened 2020-09-25
ju/ednx/LD-9 Remove preview and get urls from configuration_helpers

andrey-canon:
## Description
LD-9 requires eox-tenant in order to override SiteConfigurations https://github.com/eduNEXT/eox-tenant/pull/104

### DIFF
--- a/common/djangoapps/util/course.py
+++ b/common/djangoapps/util/course.py
@@ -54,8 +54,14 @@ def get_link_for_about_page(course):
     elif settings.FEATURES.get('ENABLE_MKTG_SITE') and getattr(course, 'marketing_url', None):
         course_about_url = course.marketing_url
     else:
+        about_base = configuration_helpers.get_value_for_org(
+            course.id.org,
+            'LMS_ROOT_URL',
+            settings.LMS_ROOT_URL
+        )
+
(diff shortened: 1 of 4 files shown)


common/djangoapps/util/course.py
```diff
@@ -54,8 +54,20 @@ def get_link_for_about_page(course):
     elif settings.FEATURES.get('ENABLE_MKTG_SITE') and getattr(course, 'marketing_url', None):
         course_about_url = course.marketing_url
     else:
+        about_base = configuration_helpers.get_value_for_org(
+            course.id.org,
+            'SITE_NAME',
```
felipemontoya, line 59: If we are moving to use LMS_BASE instead of SITE_NAME, we should already do that here

felipemontoya: Actually, we can move all this into looking for LMS_ROOT_URL which already has all everything including the `https://` bit

felipemontoya (changes requested):

common/djangoapps/util/course.py
andrey-canon: @felipemontoya  done

mariajgrimaldi (approved):

PR #418, opened 2020-09-25
Update eox_tenant

andrey-canon:
## Description
Update eox_tenant, this version of tenant is not versioned since this version only include migrations  and the official release will include the features related with the migrations


felipemontoya:
I would still prefer that we do a partial version with a version number instead of the specific commit.
We can call it 3.1.pre_juniper o something like that

mariajgrimaldi (approved):

PR #422, opened 2020-09-29
Updating authentication file with bearer authentication

vad33r:
Add support to django-oauth2-provider and django-oauth-toolkit

### DIFF
--- a/openedx/core/lib/api/authentication.py
+++ b/openedx/core/lib/api/authentication.py
@@ -7,16 +7,143 @@ from oauth2_provider import models as dot_models
 from provider.oauth2 import models as dop_models
 from rest_framework.exceptions import AuthenticationFailed
 from rest_framework_oauth.authentication import OAuth2Authentication
+from rest_framework.authentication import BaseAuthentication, get_authorization_header  # pylint: disable=ungrouped-imports
+from edx_django_utils.monitoring import set_custom_metric
 
 
-OAUTH2_TOKEN_ERROR = u'token_error'
-OAUTH2_TOKEN_ERROR_EXPIRED = u'token_expired'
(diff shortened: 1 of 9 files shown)


openedx/core/lib/api/authentication.py
```diff
@@ -105,3 +136,28 @@ def _get_dot_token(self, access_token):
...
+
+    But until then, this authentication class ensures that the user is logged in,
+    but does not require that their account "is_active".
+
+    This class can be used for an OAuth2-accessible endpoint that allows users to access
+    that endpoint without having their email verified.  For example, this is used
+    for mobile endpoints.
+    """
+
+    allow_inactive_users = True
```
andrey-canon, line 163: the last empty line is missing, Why do you include this line? where `allow_inactive_users` is used ?

package-lock.json
```diff
@@ -254,7 +254,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
+      "integrity": "sha1-+PLIh60Qv2f2NPAFtph/7TF5qsg="
```
andrey-canon, line 257: This file must not be included in this pr

andrey-canon (changes requested):

openedx/core/lib/api/authentication.py
mariajgrimaldi: is this class used?

felipemontoya (approved):
Lets fix the test and write the migration path document somewhere, but I'll leave the approval already

PR #431, opened 2020-10-05
URL enconding the username in the path for the submission history URL

cocococosti:
This PR solves an issue when using the "Submission history" functionality in the LMS (from a staff account).

When the username had spaces in it, the submission history wouldn't load, generating the following error:

Screenshot from 2020-10-05 10-05-41

This error was caused because the username in the url was passed as clear text without encoding it first.

After the fix:

Screenshot from 2020-10-05 11-22-53

Thanks to @spokerman12 for spotting the error https://3.basecamp.com/3966315/buckets/15592185/todos/2960032490/

### DIFF
--- a/lms/templates/courseware/xqa_interface.html
+++ b/lms/templates/courseware/xqa_interface.html
@@ -25,7 +25,7 @@ function setup_debug(element_id, edit_link, staff_context){
             var location = $("#" + element_id + "_history_location").val();
 
             $("#" + element_id + "_history_text").load('/courses/' + "${unicode(course.id) | u}" +
-                "/submission_history/" + username + "/" + location);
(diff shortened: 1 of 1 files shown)


spokerman12 (approved):

felipemontoya (approved):
This looks good to me. Is this something than can be made a PR to upstrea?

cocococosti:
yes, I'll make the PR to upstream @felipemontoya

felipemontoya:
@cocococosti did you make the PR?
We are looking into recreating this for juniper. It will still be necessary right? Maybe not for Koa

cocococosti:
not yet, I'll do it later today. And yes, I think this will be necessary in juniper stil @felipemontoya

PR #433, opened 2020-10-08
ju/ednx/JD-1: Microsite aware key-secret pairs for oauth

mariajgrimaldi:
**Description**

This PR allows each tenant to have its own key-secret pair. 

**Changes and justification**

- Changes KEY_FIELD from `KEY_FIELDS = ('backend_name')` to `KEY_FIELDS = ('site_id', 'backend_name')`: this allows to create differents OauthProviderConfig per site.

For example: 

With  `KEY_FIELDS = ('backend_name')`, we could only have a max of active OauthProviderConfig equals the number of available backends. 
Say that we have an active OauthProviderConfig for Google and Facebook, if we add a new one for Google even with different data but with the same backend -for Google-, the old config will be updated and the new one will be the active.

Here some proof of that:

1. I had a Provider Configuration with the backend: google-oauth2 for the site .ngrok.io
Screenshot from 2020-10-13 12-09-47

2. After I added a new Config for test-cert with the same backend, this happened:
Screenshot from 2020-10-13 12-10-33

But with `KEY_FIELDS = ('site_id', 'backend_name')` we can have the same max but per site, so we can have for each site a new config but with different backends.

For this, I followed this PR: /edx/edx-platform/pull/17276

- Pass site_id to current method calls: after changing KEY_FIELDS the current will use them as lookup fields to find the current configs, so now we have to pass the site_id.

- Override provider_id to cast field: adding site_id to KEY_FIELDS makes `provider_id` to change and add site_id (which is a number). 

- Change get_setting so the key can be searched in settings: this makes it more flexible to store key/secrets, they can be stored using the OauthProviderConfig or the tenant settings.

**Consequences**
- current method override is no longer used
- THIRD_PARTY_AUTH_ENABLED_PROVIDERS is no longer used
- From what I understand, KEY_FIELDS makes unique active configurations. So, if in PROD there are two configurations with the same SITE and BACKEND the active configuration will be the latest one.
- Because we change how current is called, some tests fail. I already fixed them. 

**Data changes**
https://docs.google.com/document/d/1CMoG_YoaYIuIGLQG2kDbvMQ4k0QNpu4GXAd3bSplCxA

**Questions and possible answers**

- why not use site_id and slug instead of backend? Well, here's the reason:
Here https://github.com/eduNEXT/edunext-platform/blob/9a69118b22f3e62eb15fd9fc274a22629d2da971/common/djangoapps/third_party_auth/provider.py#L34 instead of searching for backend_name, that the max is the number of backends available, we would need to look for slugs that normally exceed the number of backends available

LMS logs during the login process using Google:
Screenshot from 2020-10-14 08-50-16

### DIFF
--- a/common/djangoapps/third_party_auth/__init__.py
+++ b/common/djangoapps/third_party_auth/__init__.py
@@ -14,5 +14,6 @@ def is_enabled():
 
     return configuration_helpers.get_value(
         "ENABLE_THIRD_PARTY_AUTH",
-        settings.FEATURES.get("ENABLE_THIRD_PARTY_AUTH")
+        # This forces the module to be enabled on a per tenant basis
+        settings.FEATURES.get("ENABLE_THIRD_PARTY_AUTH_FOR_TEST", False)
     )
--- a/common/djangoapps/third_party_auth/models.py
+++ b/common/djangoapps/third_party_auth/models.py
@@ -348,7 +348,9 @@ class OAuth2ProviderConfig(ProviderConfig):
     # example:
     # class SecondOpenIDProvider(OpenIDAuth):
     #   name = "second-openId-provider"
-    KEY_FIELDS = ('backend_name',)
+    # eduNEXT: added site_id to KEY_FIELDS so we can have an active OAuth2ProviderConfig per site for each backend.
+    # This will change the calls to the method current, now it needs to be called passing site_id and backend_name.
+    KEY_FIELDS = ('site_id', 'backend_name')
     prefix = 'oa2'
     backend_name = models.CharField(
         max_length=50, blank=False, db_index=True,
@@ -376,6 +378,25 @@ class OAuth2ProviderConfig(ProviderConfig):
         app_label = "third_party_auth"
         verbose_name = u"Provider Configuration (OAuth)"
         verbose_name_plural = verbose_name
+
+    @classmethod
+    def current(cls, *args):
+        """
+        Get the current config model for the provider according to the given backend and the current
+        site.
+        """
+        site_id = Site.objects.get_current(get_current_request()).id
+        return super(OAuth2ProviderConfig, cls).current(site_id, *args)
+
+    @property
+    def provider_id(self):
+        """
+        Unique string key identifying this provider. Must be URL and css class friendly.
+
+        eduNEXT: override method to cast site_id field.
+        """
+        assert self.prefix is not None
+        return "-".join((self.prefix, ) + tuple(str(getattr(self, field)) for field in self.KEY_FIELDS))
 
     def clean(self):
         """ Standardize and validate fields """
--- a/openedx/core/djangoapps/auth_exchange/tests/test_forms.py
+++ b/openedx/core/djangoapps/auth_exchange/tests/test_forms.py
@@ -3,7 +3,7 @@
 Tests for OAuth token exchange forms
 """
 
-
+import os
 import unittest
 
 import httpretty
@@ -53,7 +53,7 @@ class AccessTokenExchangeFormTest(AccessTokenExchangeTestMixin):
 
 
 # This is necessary because cms does not implement third party auth
-@unittest.skipUnless(TPA_FEATURE_ENABLED, TPA_FEATURES_KEY + " not enabled")
+@unittest.skipIf(os.environ.get("CIRCLECI") == 'true', "Skip this test in Circle CI.")
 @httpretty.activate
 class DOTAccessTokenExchangeFormTestFacebook(
         DOTAdapterMixin,
@@ -69,7 +69,7 @@ class DOTAccessTokenExchangeFormTestFacebook(
 
 
 # This is necessary because cms does not implement third party auth
-@unittest.skipUnless(TPA_FEATURE_ENABLED, TPA_FEATURES_KEY + " not enabled")
+@unittest.skipIf(os.environ.get("CIRCLECI") == 'true', "Skip this test in Circle CI.")
 @httpretty.activate
 class DOTAccessTokenExchangeFormTestGoogle(
         DOTAdapterMixin,
--- a/common/djangoapps/third_party_auth/api/tests/test_views.py
+++ b/common/djangoapps/third_party_auth/api/tests/test_views.py
@@ -98,7 +98,7 @@ class UserViewsMixin(object):
             return []
         return [
             {
-                "provider_id": "oa2-google-oauth2",
+                "provider_id": "oa2-1-google-oauth2",
                 "name": "Google",
                 "remote_id": "{}@gmail.com".format(username),
             },
@@ -384,6 +384,6 @@ class TestThirdPartyAuthUserStatusView(ThirdPartyAuthTestMixin, APITestCase):
                 "disconnect_url": "/auth/disconnect/google-oauth2/?",
                 "connect_url": "/auth/login/google-oauth2/?auth_entry=account_settings&next=%2Faccount%2Fsettings",
                 "connected": False,
-                "id": "oa2-google-oauth2"
+                "id": "oa2-1-google-oauth2"
             }]
         )
--- a/common/djangoapps/third_party_auth/tests/specs/test_google.py
+++ b/common/djangoapps/third_party_auth/tests/specs/test_google.py
@@ -87,7 +87,7 @@ class GoogleOauth2IntegrationTest(base.Oauth2IntegrationTest):
         self.assertEqual(data_parsed, {
             'auth_entry': 'custom1',
             'backend_name': 'google-oauth2',
-            'provider_id': 'oa2-google-oauth2',
+            'provider_id': 'oa2-1-google-oauth2',
             'user_details': {
                 'username': 'user',
                 'email': 'user@email.com',
--- a/common/djangoapps/third_party_auth/tests/test_provider.py
+++ b/common/djangoapps/third_party_auth/tests/test_provider.py
@@ -113,12 +113,11 @@ class RegistryTest(testutil.TestCase):
         self.assertNotIn(no_log_in_provider.provider_id, provider_ids)
         self.assertIn(normal_provider.provider_id, provider_ids)
 
-    def test_tpa_hint_provider_displayed_for_login(self):
+    def test_tpa_hint_hidden_provider_displayed_for_login(self):
         """
         Tests to ensure that an enabled-but-not-visible provider is presented
         for use in the UI when the "tpa_hint" parameter is specified
         """
-
         # A hidden provider should be accessible with tpa_hint (this is the main case)
         hidden_provider = self.configure_google_provider(visible=False, enabled=True)
         provider_ids = [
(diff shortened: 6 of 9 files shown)


common/djangoapps/third_party_auth/__init__.py
```diff
@@ -14,5 +14,6 @@ def is_enabled():
 
     return configuration_helpers.get_value(
         "ENABLE_THIRD_PARTY_AUTH",
-        settings.FEATURES.get("ENABLE_THIRD_PARTY_AUTH")
+        # This forces the module to be enabled on a per microsite basis
```
felipemontoya, line 17:
```suggestion
        # This forces the module to be enabled on a per tenant basis
```

common/djangoapps/third_party_auth/models.py
```diff
@@ -348,7 +348,8 @@ class OAuth2ProviderConfig(ProviderConfig):
     # example:
     # class SecondOpenIDProvider(OpenIDAuth):
     #   name = "second-openId-provider"
-    KEY_FIELDS = ('backend_name',)
+    # eduNEXT: we keying the provider config by slug to make it more site aware.
+    KEY_FIELDS = ('slug',)
```
felipemontoya, line 352: I like this idea, but don't quite understand the implications.

Does this change the URL that acts as the entry point for the auth process?

common/djangoapps/third_party_auth/models.py
```diff
@@ -377,6 +378,29 @@ class Meta(object):
...
+        }
+        """
+        enabled_providers = getattr(
+            settings,
+            'THIRD_PARTY_AUTH_ENABLED_PROVIDERS',
+            {},
+        )
+        provider_slug = enabled_providers.get(args[0])
+        if provider_slug:
+            return super(OAuth2ProviderConfig, cls).current(provider_slug)
```
felipemontoya, line 399: Is this super().current method the one that fetches all the models and iterates through them?

felipemontoya: It might already be time to refactor that so that it does not run a for on top of all models

felipemontoya (changes requested):
I would like to understand the consequences of the key change as well as the data changes that we would need to do in the PROD data for this to work with the current clients

common/djangoapps/third_party_auth/models.py
mariajgrimaldi: I changed this from `backend_name` to `slug` because when we do this: 

https://github.com/eduNEXT/edunext-platform/blob/e86c68ad861328b04497b6457e81adbbcbd5ca92/common/djangoapps/third_party_auth/models.py#L399

The method `current` (from super) will look for the ProviderConfig using the slug field:
https://github.com/edx/django-config-models/blob/8ef133d808ff7f3efed325e74c85804b39b1436c/config_models/models.py#L139

and this filter:
https://github.com/edx/django-config-models/blob/8ef133d808ff7f3efed325e74c85804b39b1436c/config_models/models.py#L141

will be: `filter(slug=provider_slug)`

If I didn't do that, the filter would be: `filter(backend_name=provider_slug)` and then `current` would return a temporary object (not the real config)

In fact, `KEY_FIELDS = ('slug',)`  is the default as stated in: https://github.com/eduNEXT/edunext-platform/blob/e86c68ad861328b04497b6457e81adbbcbd5ca92/common/djangoapps/third_party_auth/models.py#L92

edX changed it to backend_name here:
https://github.com/edx/edx-platform/pull/20070

mariajgrimaldi: how do I check whether

> Does this change the URL that acts as the entry point for the auth process?

if it does not even work? 

For example, when I started testing with Google the button didn't even appear.

common/djangoapps/third_party_auth/models.py
mariajgrimaldi: For what I've searched this is the current method:
https://github.com/edx/django-config-models/blob/8ef133d808ff7f3efed325e74c85804b39b1436c/config_models/models.py#L128

Do you mean this line?
https://github.com/edx/django-config-models/blob/8ef133d808ff7f3efed325e74c85804b39b1436c/config_models/models.py#L141

common/djangoapps/third_party_auth/__init__.py
```diff
@@ -14,5 +14,6 @@ def is_enabled():
 
     return configuration_helpers.get_value(
         "ENABLE_THIRD_PARTY_AUTH",
-        settings.FEATURES.get("ENABLE_THIRD_PARTY_AUTH")
+        # This forces the module to be enabled on a per tenant basis
+        settings.FEATURES.get("ENABLE_THIRD_PARTY_AUTH_FOR_TEST", False)
```
andrey-canon, line 18: why is this necessary, there is a consequence if this returns None instead of False ?

common/djangoapps/third_party_auth/identityserver3.py
```diff
@@ -81,4 +81,8 @@ def get_config(self):
     @cached_property
     def _id3_config(self):
         from .models import OAuth2ProviderConfig
-        return OAuth2ProviderConfig.current("identityServer3")
+        from django.contrib.sites.models import Site
```
andrey-canon, line 84: why are you importing this inside the method ??

common/djangoapps/third_party_auth/identityserver3.py
```diff
@@ -81,4 +81,8 @@ def get_config(self):
...
+        from openedx.core.djangoapps.theming.helpers import get_current_request
```
andrey-canon, line 85: have you tried with `get_current_site` https://github.com/eduNEXT/edunext-platform/blob/ednx/juniper%2Bedunext/openedx/core/djangoapps/theming/helpers.py#L155

common/djangoapps/third_party_auth/strategy.py
```diff
@@ -30,7 +31,8 @@ def setting(self, name, default=None, backend=None):
             setting 'name' is configured via LTIProviderConfig.
         """
         if isinstance(backend, OAuthAuth):
-            provider_config = OAuth2ProviderConfig.current(backend.name)
+            site_id = Site.objects.get_current(get_current_request()).id
```
andrey-canon, line 34: same comment as above

openedx/core/djangoapps/auth_exchange/tests/test_forms.py
```diff
@@ -53,7 +53,7 @@ def _assert_success(self, data, expected_scopes):
 
 
 # This is necessary because cms does not implement third party auth
-@unittest.skipUnless(TPA_FEATURE_ENABLED, TPA_FEATURES_KEY + " not enabled")
+@unittest.skipIf(os.environ.get("CIRCLECI") == 'true', "Skip this test in Circle CI.")
```
andrey-canon, line 56: why skip ?:(

mariajgrimaldi: These tests pass locally but fail on CIRCLE CI.  I couldn't find an explanation, here the CIRCLE CI output: 
https://circleci.com/api/v1.1/project/github/eduNEXT/edunext-platform/7614/output/107/0?file=true&allocation-id=5f8eda49ee5df3231c3e5a90-0-build%2F5C7447E4

common/djangoapps/third_party_auth/__init__.py
mariajgrimaldi: I cherry-picked this from ir/ednx/JD-1. I'm gonna do some tests to check if there's a difference

common/djangoapps/third_party_auth/identityserver3.py
mariajgrimaldi: the problem is that if there's no request, then the method returns None instead of a default

common/djangoapps/third_party_auth/__init__.py
mariajgrimaldi: do you know why is this change? I can't figure it out. @felipemontoya

andrey-canon (approved):

mariajgrimaldi:
do you agree to merge this? @felipemontoya

common/djangoapps/third_party_auth/__init__.py
felipemontoya: I remember something about this needing to be turned on when the server starts (aka the yml file) and then be turned off for all the tenants which dont have a need for it.

common/djangoapps/third_party_auth/identityserver3.py
```diff
@@ -81,4 +84,6 @@ def get_config(self):
     @cached_property
     def _id3_config(self):
         from .models import OAuth2ProviderConfig
-        return OAuth2ProviderConfig.current("identityServer3")
+
+        site_id = Site.objects.get_current(get_current_request()).id
+        return OAuth2ProviderConfig.current(site_id, "identityServer3")
```
felipemontoya, line 89: In my mind it makes more sense to add the `site_id = Site.objects.get_current(get_current_request()).id` line inside the `.current` method of the `OAuth2ProviderConfig` class. This way we don't have to alter the signature of the function all over the codebase.

common/djangoapps/third_party_auth/models.py
```diff
@@ -377,6 +379,16 @@ class Meta(object):
         verbose_name = u"Provider Configuration (OAuth)"
         verbose_name_plural = verbose_name
 
+    @property
+    def provider_id(self):
+        """
+        Unique string key identifying this provider. Must be URL and css class friendly.
```
felipemontoya, line 385: If this must be URL and css class friendly, this means we must make it lowercase and remove spaces and maybe other chars. Then urlencode it. Why don't we force this to happen in this method?

common/djangoapps/third_party_auth/provider.py
```diff
@@ -30,9 +29,10 @@ def _enabled_providers(cls):
         Helper method that returns a generator used to iterate over all providers
         of the current site.
         """
+        site_id = Site.objects.get_current(get_current_request()).id
```
felipemontoya, line 32: All this edits would go away if we did the `current` mod

felipemontoya (changes requested):
Maybe I missed something or we already tried and something was not working, but in this particular point I think we should consider again modifying the `.current` method.

common/djangoapps/third_party_auth/identityserver3.py
mariajgrimaldi: OK, I made this change and for what I tested it seems to work.

common/djangoapps/third_party_auth/models.py
mariajgrimaldi: I don't see why this is necessary, given that the provider_id format is: `oa2-site_id-backend_name`. Where `site_1` is a number and `backend_name` is one of these: ['dummy', 'google-oauth2', 'linkedin-oauth2', 'facebook', 'azuread-oauth2', 'identityServer3']
It's already URL and CSS friendly, isn't it?

Unless we assume that backend_name can be any string.

felipemontoya (approved):

PR #438, opened 2020-10-15
ju/ednx/FFI-3 Tenant aware APIs

andrey-canon:
## Description
Include filter for enrollments

### DIFF
--- a/common/djangoapps/student/models.py
+++ b/common/djangoapps/student/models.py
@@ -44,6 +44,8 @@ from django.utils.translation import ugettext_noop
 from django_countries.fields import CountryField
 from edx_django_utils.cache import RequestCache
 from edx_rest_api_client.exceptions import SlumberBaseException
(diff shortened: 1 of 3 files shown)


mariajgrimaldi:
did you check https://github.com/eduNEXT/eox-tenant/pull/49 ? 
is it better to solve this problem here or there (in eox-tenant)?

andrey-canon:
> did you check [eduNEXT/eox-tenant#49](https://github.com/eduNEXT/eox-tenant/pull/49) ? I took that from the section "Future work" (Feature Doc)
> is it better to solve this problem here or there (in eox-tenant)?
> I'm curious

@mariajgrimaldi 
it's better to solve that in eox-tenant, the problem is that right now we don't enough time, eox-tenant has changed a lot, that pr must be refactored and I don't know if that works in productive.

mariajgrimaldi (approved):

PR #440, opened 2020-10-19
JD-1 Forward Compatibility

mariajgrimaldi:
Now the KEY and SECRET for Oauth can be found in:
1. The pair key-secret can be found in the tenant settings as:
```
{
  SOCIAL_AUTH_OAUTH_SECRETS: {
     KEY: <KEY>,
     SECRET: <SECRET>
  }
}
```
2. The pair key-secret can be found using the OauthProviderConfig model
3. The secret can be found in the tenant settings as:
```
{
  SOCIAL_AUTH_OAUTH_SECRETS: <SECRET>
}
```
4. The secret can be found using MICROSITE_SECRETS variable

After the juniper migration, the goal is to use only 2 and 3.

### DIFF
--- a/common/djangoapps/third_party_auth/models.py
+++ b/common/djangoapps/third_party_auth/models.py
@@ -18,7 +18,7 @@ from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 from provider.oauth2.models import Client
 from provider.utils import long_token
-from six import text_type
+from six import string_types, text_type
 from social_core.backends.base import BaseAuth
 from social_core.backends.oauth import OAuthAuth
 from social_core.backends.saml import SAMLAuth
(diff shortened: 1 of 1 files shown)


common/djangoapps/third_party_auth/models.py
```diff
@@ -385,23 +385,25 @@ def clean(self):
...
-                secrets = settings.MICROSITE_SECRETS.get(site_key, {}).get('SOCIAL_AUTH_OAUTH_SECRETS', {})
-                return current.get('SECRET', secrets.get(self.backend_name))
+        tenant_oauth = getattr(settings, 'SOCIAL_AUTH_OAUTH_SECRETS', {})
 
         if name == "KEY":
-            return self.key
+            if self.key:
+                return self.key
+            # In case we have KEY in tenant settings:
+            return tenant_oauth.get(self.backend_name, {}).get('KEY')
```
andrey-canon, line 394:
```suggestion
            # In case we have KEY in tenant settings:
            return self.key if self.key else tenant_oauth.get(self.backend_name, {}).get('KEY')
```

common/djangoapps/third_party_auth/models.py
```diff
@@ -385,23 +385,25 @@ def clean(self):
...
-            # To allow instances to avoid storing secrets in the DB, the secret can also be set via Django:
-            return getattr(settings, 'SOCIAL_AUTH_OAUTH_SECRETS', {}).get(self.backend_name, '')
+            # In case we don't have secrets in the model:
+            site_key = getattr(settings, 'microsite_config_key', '')
+            secrets = getattr(settings, MICROSITE_SECRETS, {}).get(site_key, {}).get('SOCIAL_AUTH_OAUTH_SECRETS', {})
+            # In case we have BACKEND_NAME: <SECRET> or BACKEND_NAME: {SECRET: <SECRET>} or MICROSITE_SECRETS
+            current = tenant_oauth.get(self.backend_name, {})
+            return current if isinstance(current, string_types) else current.get(
+                'SECRET',
+                secrets.get(self.backend_name)
```
andrey-canon, line 405: when there is no `self.backend_name`, is  `None` a valid default value ???

andrey-canon (changes requested):

common/djangoapps/third_party_auth/models.py
felipemontoya: Both of this 2 versions put the higher precedence on the `self.key`. I don't think that is where is at right now. The current value of self.key might be incorrect. I suggest we turn it around so that the value of the settings is read first.

common/djangoapps/third_party_auth/models.py
mariajgrimaldi: According to this line: https://github.com/eduNEXT/edunext-platform/blob/61bba889731d8ecdb47eead24e6c2a08a68320bc/common/djangoapps/third_party_auth/models.py#L404

It should be `""`
I already changed it

common/djangoapps/third_party_auth/models.py
mariajgrimaldi: Yes, you're right considering that self.key might be incorrect. I'll change it so we check first the settings

felipemontoya (approved):

common/djangoapps/third_party_auth/models.py
```diff
@@ -385,23 +385,26 @@ def clean(self):
...
-            # To allow instances to avoid storing secrets in the DB, the secret can also be set via Django:
-            return getattr(settings, 'SOCIAL_AUTH_OAUTH_SECRETS', {}).get(self.backend_name, '')
+            # In case we don't have secrets in the model:
+            site_key = getattr(settings, 'microsite_config_key', '')
+            secrets = getattr(settings, 'MICROSITE_SECRETS', {}).get(site_key, {}).get('SOCIAL_AUTH_OAUTH_SECRETS', {})
+            # In case we have BACKEND_NAME: <SECRET> or BACKEND_NAME: {SECRET: <SECRET>} or MICROSITE_SECRETS
+            current = tenant_oauth.get(self.backend_name, {})
+            return current if isinstance(current, string_types) else current.get(
+                'SECRET',
+                secrets.get(self.backend_name, '')
```
andrey-canon, line 406: trailing comma

PR #443, opened 2020-10-23
ju/ednx/JU-6: Hooks support

mariajgrimaldi:
**Description:**

Adds support to the first implementation of the [Hooks](https://drive.google.com/a/edunext.co/open?id=1iep7Hcac7oylmpLHqnPnvOeywCqyiIRLhPIssHMNpwQ) concept in Juniper. 

The implementation consists of the pre_enrollment hook, that if enabled, sends a signal that executes a pre-defined action.

**Tests:**

- [x] preenrollment_hook:  test that the signal sent during the enrollment process triggers an action. In this test, the action used was the one defined in eox-taylor.


andrey-canon (approved):
Fix quality tests before merging

PR #448, opened 2020-11-02
ju/ednx/JU-3: Fix CertificateGenerated.save proxy

mariajgrimaldi:
PR related to https://github.com/eduNEXT/edunext-platform/pull/372 

**Description and justification**

This PR makes CertificateGenerated.save work with or without proxy.

Chaging _super_ from `super(GeneratedCertificate, self)` -python 2.7- to `super()` -python 3- ([reference](https://www.python.org/dev/peps/pep-3135/))   fixes the following error when using `TenantGeneratedCertificateProxy`:

```
super(GeneratedCertificate, self).save(*args, **kwargs)
TypeError: super(type, obj): obj must be an instance or subtype of type
```

This error was raised because `type` or GeneratedCertificate was  `<class 'eox_tenant.tenant_wise.proxies.TenantGeneratedCertificateProxy'>` instead of` <class 'lms.djangoapps.certificates.models.GeneratedCertificate'>`. 

After the super().save(*args, **kwargs) finishes,  the respective signals are sent with `sender=self.__class__ `(`or <class 'lms.djangoapps.certificates.models.GeneratedCertificate'>`)
 
**Tests**

Tested adding a new GeneratedCertificate using django-admin.

Before JU-3:
Screenshot from 2020-11-02 12-59-07

After:
Screenshot from 2020-11-02 12-59-56

and  `self.__class__ == <class 'lms.djangoapps.certificates.models.GeneratedCertificate'>`

The COURSE_CERT_CHANGED receivers -`handle_course_cert_changed` and `handle_cert_change`- were tested to check the correct signal functionality.

### DIFF
--- a/lms/djangoapps/certificates/models.py
+++ b/lms/djangoapps/certificates/models.py
@@ -401,7 +401,7 @@ class GeneratedCertificate(models.Model):
         signal iff we are saving a record of a learner passing the course.
         As well as the COURSE_CERT_CHANGED for any save event.
         """
-        super(GeneratedCertificate, self).save(*args, **kwargs)
+        super().save(*args, **kwargs)
         COURSE_CERT_CHANGED.send_robust(
             sender=self.__class__,
             user=self.user,


andrey-canon (approved):

mariajgrimaldi:
Can you check this? @felipemontoya 

You left this comment in Pull 372 that says:

`This change can be evaluated during the upgrade to juniper and python3 maybe it is not necessary then.`

this change -made in this PR- was what you meant?

felipemontoya:
This looks good to me @mariajgrimaldi. From my comment in #372, actually looks like it is still necessary, but it is much more simple this time.

felipemontoya (approved):

felipemontoya:
Also, I think you did a good research of the underlying issue. :+1:

PR #459, opened 2020-11-24
ju/ednx/BC-19_P7: Fix lti-consumer versions and duplicates

mariajgrimaldi:
Removed lti consumer from requirements/edx and downgraded lti-consumer to 1.2.6 same used in edx. 
Our version is v1.2.6.1, the same as edx but with `expose fullname` patch.

NOTE: the changes made in base.txt were made by `make edunext-upgrade` after changing lti-consumer version


felipemontoya (approved):
I reviewed the versions and this is ready for merge

PR #460, opened 2020-11-24
Adding new translations to edunext-platform

vad33r:
Adding translations for the juniper-release + master of 2020-11-05 and Hindi


conf/locale/config.yaml
```diff
@@ -46,46 +46,42 @@ locales:
...
+    # - id  # Indonesian
     - it_IT  # Italian (Italy)
     - ja_JP  # Japanese (Japan)
-    - ka  # Georgian
     # - kk_KZ  # Kazakh (Kazakhstan)
     # - km_KH  # Khmer (Cambodia)
     # - kn  # Kannada
     - ko_KR  # Korean (Korea)
-    - lt_LT  # Lithuanian (Lithuania)
-    - lv  # Latvian
```
felipemontoya, line 58: The standard we are using is to comment out the languages that we are not using. Not removing them from the list

conf/locale/config.yaml
```diff
@@ -132,7 +128,7 @@ ignore_dirs:
     - src/done-xblock
     - src/edx-jsme
     - src/parse-rest
-    - src/geoip2
```
felipemontoya, line 135: This does not look to me like something we should be changing here. Is this an oversight or do we really need it?

conf/locale/config.yaml
```diff
@@ -161,10 +157,6 @@ segment:
     djangojs-partial.po:
         djangojs-studio.po:
             - cms/*
-        djangojs-account-settings-view.po:
```
felipemontoya, line 164: This does not look to me like something we should be changing here. Is this an oversight or do we really need it?

conf/locale/config.yaml
```diff
@@ -184,6 +176,5 @@ generate_merge:
     djangojs.po:
         - djangojs-partial.po
         - djangojs-studio.po
-        - djangojs-account-settings-view.po
```
felipemontoya, line 187: This does not look to me like something we should be changing here. Is this an oversight or do we really need it?

felipemontoya (changes requested):
Some minor changes in the config.yaml. The rest looks good

conf/locale/config.yaml
vad33r: Ok, i will fixed it and search for another deleted languages

conf/locale/config.yaml
vad33r: I think this change is an oversight because of the compiled

felipemontoya (approved):
Good, this looks like it is ready to merge

PR #462, opened 2020-11-26
ju/ednx/BC-19_P8: Fix #egg in lti-consumer/scorm xblock 

mariajgrimaldi:
Changed #egg in github.in so it's the same as the one used by edX -LTI CONSUMER-  or by edunext/ironwood release -SCORM-


morenol (approved):
lgtm

PR #466, opened 2020-12-09
Fixed printing syntax error

mariajgrimaldi:
This PR fixes the printing error -in python 3.5- in problem_with_hint.yaml that affected the creation of problems with adaptive hints

### DIFF
--- a/common/lib/xmodule/xmodule/templates/problem/problem_with_hint.yaml
+++ b/common/lib/xmodule/xmodule/templates/problem/problem_with_hint.yaml
@@ -11,7 +11,7 @@ data: |
             <customresponse cfn="test_str" expect="python">
             <script type="text/python" system_path="python_lib">
     def test_str(expect, ans):
-      print expect, ans
+      print(expect, ans)
       ans = ans.strip("'")
       ans = ans.strip('"')
       return expect == ans.lower()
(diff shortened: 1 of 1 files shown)


felipemontoya:
@mariajgrimaldi althought this looks good to me, I do have a few questions.

This is only a template that outputs some example code right? Is the example code meant to be run in the codejailed code?

@ericfab179 are we running the codejail still on python2.7?

felipemontoya:
@mariajgrimaldi if we merge this, it would be important to also make an OSPR request with the same changes

mariajgrimaldi:
Hey, @felipemontoya 

1. This is only a template that outputs some example code right?
     Yes, that's what I know.

2. Is the example code meant to be run in the codejailed code?
Yes. The code is executed using this method:
https://github.com/eduNEXT/edunext-platform/blob/59a39c67ca1e886e4ec8c9f6e669ac4ba58cef19/common/lib/capa/capa/capa_problem.py#L916

Defined here:
https://github.com/eduNEXT/edunext-platform/blob/59a39c67ca1e886e4ec8c9f6e669ac4ba58cef19/common/lib/capa/capa/safe_exec/safe_exec.py#L82

And that method uses `safe_exec` from `codejail`. Is that what you meant?

felipemontoya:
Then it is more important than ever to know if codejail is running on python2.7 or 3.5. cc @jfavellar90 and @ericfab179

mariajgrimaldi:
Hey @felipemontoya I added `sys.version` to the _hint_ and this was what I got:
Screenshot from 2020-12-10 09-33-27

felipemontoya:
Yep. I talked to @jfavellar90 and he confirmed that we are using 3.5 but we will be moving to 2.7 before the migration

mariajgrimaldi:
> Yep. I talked to @jfavellar90 and he confirmed that we are using 3.5 but we will be moving to 2.7 before the migration

should I close this then?

PR #475, opened 2020-12-30
Cache response from yt_video_metadata view

magajh:
This PR allows caching the response from the youtube video metadata view.

### DIFF
--- a/lms/djangoapps/courseware/views/views.py
+++ b/lms/djangoapps/courseware/views/views.py
@@ -298,8 +298,13 @@ def load_metadata_from_youtube(video_id, request):
     This method is used via the standalone /courses/yt_video_metadata REST API
     endpoint, or via the video XBlock as a its 'yt_video_metadata' handler.
     """
-    metadata = {}
+    cache_key = "yt_video_metadata_cache.{video}".format(video=video_id)
+    metadata = cache.get(cache_key, {})
     status_code = 500
+
+    if metadata:
+        return metadata, 200
+
     if video_id and settings.YOUTUBE_API_KEY and settings.YOUTUBE_API_KEY != 'PUT_YOUR_API_KEY_HERE':
         yt_api_key = settings.YOUTUBE_API_KEY
         yt_metadata_url = settings.YOUTUBE['METADATA_URL']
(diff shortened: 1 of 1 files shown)


mariajgrimaldi (approved):

felipemontoya (commented):
I was looking at the cache_page function at:
https://docs.djangoproject.com/en/3.1/topics/cache/

But it is not clear for me that the different IDs of different videos will be cached separately.

Are you sure it works in this scenario?

lms/djangoapps/courseware/views/views.py
```diff
@@ -298,40 +298,50 @@ def load_metadata_from_youtube(video_id, request):
...
-                        metadata = res_json
-                    else:
-                        logging.warning(u'Unable to find the items in response. Following response '
+    cache_key = "yt_video_metadata_cache.{video}".format(video=video_id)
+    metadata = cache.get(cache_key, {})
+    status_code = 200
+
+    if not metadata:  # pylint: disable=too-many-nested-blocks
+        status_code = 500
+        if video_id and settings.YOUTUBE_API_KEY and settings.YOUTUBE_API_KEY != 'PUT_YOUR_API_KEY_HERE':
```
mariajgrimaldi, line 307: Why not use getattr?

lms/djangoapps/courseware/views/views.py
```diff
@@ -298,40 +298,50 @@ def load_metadata_from_youtube(video_id, request):
...
-                    res_json = res.json()
-                    if res_json.get('items', []):
-                        metadata = res_json
-                    else:
-                        logging.warning(u'Unable to find the items in response. Following response '
+    cache_key = "yt_video_metadata_cache.{video}".format(video=video_id)
+    metadata = cache.get(cache_key, {})
+    status_code = 200
+
+    if not metadata:  # pylint: disable=too-many-nested-blocks
```
mariajgrimaldi, line 305: maybe you could remove `disable=too-many-nested-blocks` by adding 
if metadata return metadata, status_code
... rest of code
return metadata, status_code

???

magajh: I changed this. Thank you

felipemontoya (approved):

PR #477, opened 2020-12-30
Avoid duplicate experimetdata creation

andrey-canon:
## Description 
This change comes from https://github.com/edx/edx-platform/pull/25863,  I'm not sure if this will fix the current errors since I couldn't replicate the error however this improves the code readability and doesn't generate conflict with the upstream version.


felipemontoya:
Way better to use get_or_create

felipemontoya (approved):

felipemontoya:
For the tests to pass, you need to rebase the firefox fix

PR #479, opened 2020-12-30
Add extended profile fields to the jwt payload.

andrey-canon:
## Description
Add extended profile fields to the jwt payload in order to share any custom field data in the jwt response.

### DIFF
--- a/openedx/core/djangoapps/oauth_dispatch/jwt.py
+++ b/openedx/core/djangoapps/oauth_dispatch/jwt.py
@@ -9,7 +9,7 @@ from edx_django_utils.monitoring import set_custom_metric
 from edx_rbac.utils import create_role_auth_claim_for_user
 from jwkest import jwk
 from jwkest.jws import JWS
-
+from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
 from student.models import UserProfile, anonymous_id_for_user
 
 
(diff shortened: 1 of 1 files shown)


mariajgrimaldi (changes requested):
I checked CMS tests and found this
```
payload = {'aud': 'video_pipeline_client_id', 'email': 'robot+test+639@edx.org', 'email_verified': True, 'exp': 1609373083, ...}
user = <User: cms_video_pipeline_service_user>

    def _attach_profile_claim(payload, user):
        """Add the profile claim details to the JWT payload."""
        try:
            # Some users (e.g., service users) may not have user profiles.
            profile = UserProfile.objects.get(user=user)
            name = profile.name
            fields = configuration_helpers.get_value('extended_profile_fields', [])
    
            if fields:
                extended_profile_fields = {key: value for key, value in json.loads(profile.meta).items() if key in fields}
    
        except UserProfile.DoesNotExist:
            name = None
            extended_profile_fields = None
    
        payload.update({
            'name': name,
            'family_name': user.last_name,
            'given_name': user.first_name,
            'administrator': user.is_staff,
            'superuser': user.is_superuser,
>           'extended_profile_fields': extended_profile_fields,
        })
E       UnboundLocalError: local variable 'extended_profile_fields' referenced before assignment

openedx/core/djangoapps/oauth_dispatch/jwt.py:200: UnboundLocalError
```

mariajgrimaldi (approved):

openedx/core/djangoapps/oauth_dispatch/jwt.py
```diff
@@ -178,9 +178,17 @@ def _attach_email_claim(payload, user):
 
 def _attach_profile_claim(payload, user):
     """Add the profile claim details to the JWT payload."""
+    extended_profile_fields = None
+
     try:
         # Some users (e.g., service users) may not have user profiles.
-        name = UserProfile.objects.get(user=user).name
+        profile = UserProfile.objects.get(user=user)
+        name = profile.name
```
felipemontoya, line 186: Can we have a switch to turn this on or off?

I'm worried that this might be too heavy for the JWT and we have no way to turn it off on the fly.

Question, is this necessary only for some sites? if so, should we leave the switch be Off by default?

By switch I mean setting that we can alter from the tenant.

felipemontoya (changes requested):

openedx/core/djangoapps/oauth_dispatch/jwt.py
```diff
@@ -178,9 +178,17 @@ def _attach_email_claim(payload, user):
...
+        fields = configuration_helpers.get_value('extended_profile_fields', [])
+
+        if fields and configuration_helpers.get_value('JWT_EXTENDED_PROFILE', False):
```
andrey-canon, line 189: @felipemontoya  done in this line

felipemontoya (approved):

PR #481, opened 2021-01-04
Added capability to ignore fields -in register form- when creating them 

mariajgrimaldi:
This PR avoids raising errors when:

1. A field is added to REGISTRATION_EXTRA_FIELDS
2. We aren't interested in creating its definition dictionary because we don't want to render the field. For that, we don't want to raise any errors to not cause noise in sentry.

This is done to cover a special case with `authorized_email` field.

Now, to ignore it we will have to add the setting` EDNX_IGNORE_REGISTER_FIELDS : ["authorized_email"]`

### DIFF
--- a/openedx/core/djangoapps/user_authn/views/registration_form.py
+++ b/openedx/core/djangoapps/user_authn/views/registration_form.py
@@ -679,6 +679,9 @@ class RegistrationFormFactory(object):
             field_name (str): Name used to get field information when creating it.
         """
         field_name = kwargs.pop("field_name")
+        if field_name in getattr(settings, "EDNX_IGNORE_REGISTER_FIELDS", []):
+            return
+
         custom_field_dict = self._get_custom_field_dict(field_name)
         if not custom_field_dict:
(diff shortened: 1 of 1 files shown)


lms/envs/common.py
```diff
@@ -3919,3 +3919,6 @@ def _make_locale_paths(settings):  # pylint: disable=missing-function-docstring
 
 #################### DELETION TEXT ###########################
 SITE_SPECIFIC_DELETION_TEXT = ''
+
+################ REGISTER FIELDS TO IGNORE ###################
+EDNX_IGNORE_REGISTER_FIELDS = ["authorized_email"]
```
andrey-canon, line 3924: why is this the default value ?

mariajgrimaldi: This is done to ignore this: https://sentry.io/organizations/edunext/issues/2117363443/?project=5181932&query=is%3Aunresolved

That field is used in the WhitelistAccess feature

felipemontoya (approved):

PR #488, opened 2021-01-21
Restrict access token by application allowed uris.

andrey-canon:
## Description 
Restrict access token by application allowed uris.


felipemontoya (changes requested):
Make this a subclass by the same name as we spoke per other channels

felipemontoya (approved):
Beautiful, squash and merge

PR #494, opened 2021-02-02
Added another format for registration field options

mariajgrimaldi:
This PR adds a new format for registration field options, it allows us to use:

Format 1:

"options": [option1, ...., optionN]

Format 2:

"options":  {"value": "name", ..., "valueN": "nameN"}

### DIFF
--- a/openedx/core/djangoapps/user_authn/views/registration_form.py
+++ b/openedx/core/djangoapps/user_authn/views/registration_form.py
@@ -690,7 +690,10 @@ class RegistrationFormFactory(object):
         # Check to convert options:
         field_options = custom_field_dict.get("options")
         if field_options:
-            field_options = [(six.text_type(option.lower()), option) for option in field_options]
+            try:
+                field_options = [(six.text_type(value.lower()), name) for value, name in field_options.items()]
+            except AttributeError:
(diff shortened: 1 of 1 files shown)


openedx/core/djangoapps/user_authn/views/registration_form.py
```diff
@@ -690,7 +690,10 @@ def _add_custom_field(self, form_desc, required=True, **kwargs):
         # Check to convert options:
         field_options = custom_field_dict.get("options")
         if field_options:
-            field_options = [(six.text_type(option.lower()), option) for option in field_options]
+            try:
+                field_options = [(six.text_type(value.lower()), name) for value, name in field_options.items()]
```
morenol, line 694: This looks good to me. I think that we should try first the old format (```field_options = [(six.text_type(value.lower()), value) for value in field_options]```) and if that fails use the new format. What do you think?

mariajgrimaldi: But how the old format would fail? :thinking:

morenol: You are right. My bad😂

morenol (approved):

PR #496, opened 2021-02-08
Add Hook for certificate creation

magajh:
This PR adds the trigger to be called after a users certificate is generated.


lms/djangoapps/certificates/models.py
```diff
@@ -423,6 +424,15 @@ def save(self, *args, **kwargs):
                 status=self.status,
             )
 
+            post_certificate_creation.send_robust(
```
morenol, line 427: I wonder which is the practical difference between this and COURSE_CERT_AWARDED.

magajh: There is not really a difference in terms of what both of them do.
This was made more to define a signal supported for the eox-hooks consumption, along with the other triggers

felipemontoya: The big difference would be in the information we send to the action. Right now we are sending the certificate id, and the url, but I wonder why are we not sending the certificate object itself?

magajh: I changed it, now It's only sending the certificate, all the other args can be taken from it

morenol (approved):
I approving this but still not sure if this will be the final way to do this

PR #497, opened 2021-02-08
Fix create users with usernames longer than 30 chars

mariajgrimaldi:
This PR changes the max length for usernames when creating users, the max was set to the maximum used by [Django](https://docs.djangoproject.com/en/3.1/ref/contrib/auth/#:~:text=150%20characters%20or%20fewer.,and%20%2D%20characters.) itself. This change affects the creation of the user by manual registration and using TPA providers like SAML.

### DIFF
--- a/common/djangoapps/student/tests/test_long_username_email.py
+++ b/common/djangoapps/student/tests/test_long_username_email.py
@@ -25,10 +25,10 @@ class TestLongUsernameEmail(TestCase):
 
     def test_long_username(self):
         """
-        Test username cannot be more than 30 characters long.
+        Test username cannot be more than 150 characters long.
         """
 
-        self.url_params['username'] = 'username' * 4
+        self.url_params['username'] = 'username' * 20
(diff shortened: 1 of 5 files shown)


felipemontoya (approved):

felipemontoya:
@mariajgrimaldi since this should be also merged upstream I'd like to clarify that this does not affect users created via SAML or other TPA. Those already bypass the 30 char limit

mariajgrimaldi:
I made that observation because of this function that was changed: [clean_username](https://github.com/eduNEXT/edunext-platform/blob/f62b8c864bedc2c9edf993150b712b08c87e6898/common/djangoapps/third_party_auth/models.py#L72) used in [get_username](https://github.com/eduNEXT/edunext-platform/blob/f62b8c864bedc2c9edf993150b712b08c87e6898/common/djangoapps/third_party_auth/pipeline.py#L852) 

clean_username used 30 chars limit, that didn't affect the username during registration?

PR #501, opened 2021-02-17
Jhony/fix roles with no course

jfavellar90:
This PR aims to solve the following issue:

The studio performance is degraded in /home and /settings/details/<course_id> pages when a user with global instructor or staff permissions over an organization (which maps to a Corse Access Role record with a defined role and org, but not course_id) tries to access them. This is because when getting the courses for a user in https://github.com/eduNEXT/edunext-platform/blob/49c83d61bc63c168d9445823e098ce1024ee880f/cms/djangoapps/contentstore/views/course.py#L466, if a Course Access Role with no course_id is found, the platform rises an exception in https://github.com/eduNEXT/edunext-platform/blob/49c83d61bc63c168d9445823e098ce1024ee880f/cms/djangoapps/contentstore/views/course.py#L482 which fallback to a method that also gets courses for the user but causes more overhead over the mongo database (https://github.com/eduNEXT/edunext-platform/blob/49c83d61bc63c168d9445823e098ce1024ee880f/cms/djangoapps/contentstore/views/course.py#L382)

This solution consists of not raising the exception when a Course Access Role does not have a course_id and just pulling all courses from organizations the user has access to globally. The CourseOverview model is used for that purpose. The exception AccessListFallback will only be raised if a Course Access Role record contains neither course_id nor org fields.

This PR was tested on stage. The loading time of these Studio pages under the conditions described above was improved

### DIFF
--- a/cms/djangoapps/contentstore/views/course.py
+++ b/cms/djangoapps/contentstore/views/course.py
@@ -477,10 +477,20 @@ def _accessible_courses_list_from_groups(request):
     courses_list = []
     course_keys = {}
 
+    user_global_orgs = set()
     for course_access in all_courses:
-        if course_access.course_id is None:
+        if course_access.course_id is not None:
+            course_keys[course_access.course_id] = course_access.course_id
+        elif course_access.org:
+            user_global_orgs.add(course_access.org)
+        else:
             raise AccessListFallback
-        course_keys[course_access.course_id] = course_access.course_id
+
+    if user_global_orgs:
+        # Getting courses from user global orgs
+        overviews = CourseOverview.get_all_courses(orgs=list(user_global_orgs))
+        overviews_course_keys = {overview.id: overview.id for overview in overviews}
+        course_keys.update(overviews_course_keys)
 
     course_keys = list(course_keys.values())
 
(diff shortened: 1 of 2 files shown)


cocococosti (approved):

cms/djangoapps/contentstore/views/course.py
```diff
@@ -477,10 +477,20 @@ def filter_ccx(course_access):
     courses_list = []
     course_keys = {}
 
+    user_global_orgs = set()
     for course_access in all_courses:
-        if course_access.course_id is None:
+        if course_access.course_id is not None:
```
mariajgrimaldi, line 482: Why not just 

`if course_access.course_id`

mariajgrimaldi (approved):

cms/djangoapps/contentstore/views/course.py
jfavellar90: Since the old code used `if course_access.course_id is None` I decided just to use the inverse validation. course_id is a special field that can be populated either with a course locator or with None.

jfavellar90:
I'll proceed with this merge since tests already ran correctly

PR #510, opened 2021-04-07
Add post register trigger

magajh:
This PR adds the post register trigger.

The trigger was provisionally implemented with a signal in order to develop the  [UPC EXED post register Hook](https://3.basecamp.com/3966315/buckets/8050706/todos/3607953105#__recording_3636632623) feature.

### DIFF
--- a/openedx/core/djangoapps/user_authn/views/register.py
+++ b/openedx/core/djangoapps/user_authn/views/register.py
@@ -57,6 +57,7 @@ from openedx.core.djangoapps.user_authn.views.registration_form import (
     RegistrationFormFactory
 )
 from openedx.core.djangoapps.waffle_utils import WaffleFlag, WaffleFlagNamespace
+from openedx.core.lib.triggers.v1 import post_register
 from student.helpers import (
     authenticate_new_user,
     create_or_set_user_attribute_created_on_site,
@@ -234,6 +235,9 @@ def create_account_with_params(request, params):
 
     # Announce registration
     REGISTER_USER.send(sender=None, user=user, registration=registration)
+
+    # This signal allows plugins to process the registration information after the account creation process
+    post_register.send_robust(sender=None, user=user)
 
     create_comments_service_user(user)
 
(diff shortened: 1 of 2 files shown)


openedx/core/djangoapps/user_authn/views/register.py
```diff
@@ -235,6 +235,9 @@ def create_account_with_params(request, params):
     # Announce registration
     REGISTER_USER.send(sender=None, user=user, registration=registration)
 
+    # This signal allows plugins to process the registration information after the account creation process
+    post_register.send_robust(sender=None, user=user, registration=registration)
```
mariajgrimaldi, line 239: LGTM considering this is the only place where user registration occurs

mariajgrimaldi (approved):

mariajgrimaldi:
Fix tests though

openedx/core/djangoapps/user_authn/views/register.py
morenol: Why don't we just use the REGISTER_USER signal that already exists and has the same parameters

magajh: @morenol I deleted the registration parameter since it's not really used and according to [this](https://github.com/edx/edx-platform/blob/452795789367fedfbfc51223d585f9b87ff400a2/docs/guides/hooks/post_register_trigger.rst) is not planned to be passed in the future as a parameter of the trigger

magajh: @felipemontoya should we use the REGISTER_USER signal that's already implemented for this hook?

felipemontoya: @magajh I see good reasons to both leave it as a new signal and use the old signal.

In favour of using the new is that once we are ready to migrate to whatever ends up in upstream:master, we can look at any usage of the signals in `openedx/core/lib/triggers/v1.py ` and we are done.

In favour of the old is to not repeat ourselves.

If you are fine with using `send` and not `send_robust`, you could create the post_register var at  `openedx/core/lib/triggers/v1.py ` and assign the value of `REGISTER_USER`. I mean, as long as this does not create a circular dependecy problem

magajh: @felipemontoya thank you

for now, I'm leaving the duplicated signal since importing `REGISTER_USER` in `openedx/core/lib/triggers/v1.py` does create a dependency problem

PR #513, opened 2021-04-27
Updating staff graded xblock to fix js bug

cocococosti:
This PR updates the version of the Staff Graded Points xblock to correct a bug when navigating a course with the xblock.

When a unit has a Staff Graded Points component, and you navigate the course as a student, the "Next"  and "Previous" button don't work. You can see it happening in prod [here](https://sitetestalbeironotdelete.edunext.io/courses/course-v1:albeiro-hawthorn-enero-2019+test-ironwood+2019_2/courseware/4ad6d1da6bf94363af4e7bfdeb22988d/f107301e00aa447083ff5302b774a191/?child=last)

This behavior was corrected in this commit https://github.com/edx/staff_graded-xblock/pull/18/commits/d8f60c789bde3552e98af520719fd7be5ddbf545 included in version 1.1

I tested it in stage [here](https://openedx-edunext-co.je.edunextstage.net/courses/course-v1:albeiro-hawthorn-enero-2019+CTR101+2021_T1/courseware/2d9aae4111d54fa0829900193e6f7f00/09e63156892c45a4b634435f449eccfd/?child=last) and the fix works correctly.

Note: I tried installing the latest version of the xblock (1.5) but it had some conflicts with Juniper.

### DIFF
--- a/requirements/edx/base.txt
+++ b/requirements/edx/base.txt
@@ -227,7 +227,7 @@ git+https://github.com/jazzband/sorl-thumbnail.git@13bedfb7d2970809eda597e3ef793
 sortedcontainers==2.1.0   # via -r requirements/edx/base.in, pdfminer.six
 soupsieve==2.0.1          # via beautifulsoup4
 sqlparse==0.3.1           # via -r requirements/edx/base.in, django
-staff-graded-xblock==0.8  # via -r requirements/edx/base.in
+staff-graded-xblock==1.1  # via -r requirements/edx/base.in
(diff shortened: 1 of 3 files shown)


mariajgrimaldi:
should we pin the version or is it not necessary?

morenol (approved):

MoisesGSalas (approved):

cocococosti:
> should we pin the version or is it not necessary?

I don't think it's necessary @mariajgrimaldi  As long as nobody runs a make upgrade or something like that. Correct? @morenol

PR #524, opened 2021-06-17
feat: allow the user to create course modes

mariajgrimaldi:
### **Description**
This PR allows the user to create course modes directly through the studio certificates view. This PR is just backend modifications, see #527 for the backbone components that form the rest of the feature.

https://3.basecamp.com/3966315/buckets/8050706/todos/3733330052

### **How to test**
Use instructions in PR #527

### DIFF
--- a/cms/djangoapps/contentstore/views/certificates.py
+++ b/cms/djangoapps/contentstore/views/certificates.py
@@ -31,6 +31,7 @@ from django.conf import settings
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponse
+from django.urls import reverse
 from django.utils.translation import ugettext as _
 from django.views.decorators.csrf import ensure_csrf_cookie
 from django.views.decorators.http import require_http_methods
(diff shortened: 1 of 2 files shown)


cms/djangoapps/contentstore/utils.py
```diff
@@ -135,6 +136,29 @@ def get_lms_link_for_item(location, preview=False):
     )
 
 
+def get_lms_course_mode_api_endpoint(course_key):
+    """
+    Returns the course modes API URL for the given course.
+    """
+    lms_root = configuration_helpers.get_value(
+        "LMS_ROOT_URL",
```
mariajgrimaldi, line 144: Question: 
I've seen how `lms_root` is obtained in other places and is using `get_value_for_org`, [here](https://github.com/eduNEXT/edunext-platform/blob/339c5f88f02f1b6d8dfc67f1119b830a3c9f3608/openedx/core/djangoapps/user_api/accounts/utils.py#L168) an example. Should we use that instead of `get_value`? I tried but I was using the same org on different sites and it failed. I couldn't get the correct lms_root. It was because of my particular configuration or we shouldn't use it (for other reasons)? @felipemontoya

felipemontoya: When I look at it today in master, get_value_for_org seems to be the preferred way.

https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/user_api/accounts/utils.py#L168

If it is not working for you I would asume it has to do with your particular config

magajh (approved):
If you tested this and it works fine then LGTM

jignaciopm (approved):
I have zero context but I see the code well

cms/djangoapps/contentstore/views/certificates.py
```diff
@@ -397,28 +398,40 @@ def certificates_list_handler(request, course_key_string):
...
                 mode.slug for mode in CourseMode.modes_for_course(
                     course_id=course.id, include_expired=True
                 ) if mode.slug != 'audit'
             ]
+            # Check whether the audit mode associated with the course exists in database
+            has_audit_mode = CourseMode.objects.filter(course_id=course.id, mode_slug='audit')
 
             has_certificate_modes = len(course_modes) > 0
 
+            enable_course_mode_creation = False
```
MoisesGSalas, line 415: can you swap how you set `enable_course_mode_creation`? I makes more sense to me that you read the feature from the setting and if the course `has_certificate_mode` you set it to false.

cms/djangoapps/contentstore/views/certificates.py
```diff
@@ -397,28 +398,40 @@ def certificates_list_handler(request, course_key_string):
                 handler_name='certificate_activation_handler',
                 course_key=course_key
             )
+            course_mode_creation_url = reverse(
+                'course_modes_api:v1:course_modes_list',
+                args=(text_type(course_key),)
```
andrey-canon, line 403: trailing comma hahaha

cms/djangoapps/contentstore/views/certificates.py
mariajgrimaldi: According to the [basecamp](https://3.basecamp.com/3966315/buckets/8050706/todos/3733330052) discussion, even when the course has audit mode, the option shouldn't be displayed. The problem is, given that audit is the default, it's ignored in `has_certificate_modes` so we can't count on it to enable the feature.

Wha I can do is switch enable_course_mode_creation = settings.FEATURES.get('ENABLE_COURSE_MODE_CREATION', False)` to this line. You're right, it makes more sense

MoisesGSalas (approved):

PR #525, opened 2021-06-24
feat: expose course modes API to studio

mariajgrimaldi:
This PR exposes the course modes API so we can create them in studio. Solves communication issues between Studio-LMS.

More context: https://3.basecamp.com/3966315/buckets/8050706/todos/3733330052

### DIFF
--- a/cms/urls.py
+++ b/cms/urls.py
@@ -175,6 +175,9 @@ urlpatterns = [
     url(r'^api/val/v0/', include('edxval.urls')),
     url(r'^api/tasks/v0/', include('user_tasks.urls')),
     url(r'^accessibility$', contentstore.views.accessibility, name='accessibility'),
+    # Course modes API for certificates generation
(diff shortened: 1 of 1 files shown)


magajh (approved):

cms/urls.py
```diff
@@ -175,6 +175,9 @@
     url(r'^api/val/v0/', include('edxval.urls')),
     url(r'^api/tasks/v0/', include('user_tasks.urls')),
     url(r'^accessibility$', contentstore.views.accessibility, name='accessibility'),
+    # Course modes API for certificates generation
```
jignaciopm, line 178: are these comments necessary?

jignaciopm (approved):

andrey-canon (approved):

cms/urls.py
mariajgrimaldi: I think so, these URLs don't belong here. So it's nice to leave a short explanation

PR #545, opened 2021-08-19
fix: change the egg of the requirement xblock pdf

MaferMazu:
Changing the xblock pdf requirement and keeping the egg (git+github.com/appsembler/pdfXBlock.git@v0.3.1 **#egg=xblock-pdf** ) in the `github.in` file caused the change to not be made correctly.

The first option to change the egg was: `xblock-pdf==0.3.1`
But when simulating productive (having an xblock already installed and installing another with the specified egg) it gave me the following warning.
Screenshot from 2021-08-18 21-27-02

And I tried simulating productive with this other egg `xblock-pdf-appsembler`:
Screenshot from 2021-08-18 21-29-39

When pip freeze is done it keeps a neutral name in both cases
Screenshot from 2021-08-18 21-30-25

The package to keep is the pdf-xblock==0.3.1. I have to tell Jhony to remove the other if it is still in prod.


mariajgrimaldi:
please rebase

MaferMazu:
Ready @mariajgrimaldi

mariajgrimaldi:
Well, let's see if it works.

mariajgrimaldi (approved):

PR #553, opened 2021-10-06
Gonzalo/githubactions unit tests

NeOneSoft:
This PR was created in order to run unit tests for eduNEXT limonero use case.

PR #550, opened 2021-10-01
[WIP] BOM-2562: Move JS Tests To Github Action (#27744)

mariajgrimaldi:
* feat: JS tests on Github Action

PR #549, opened 2021-10-01
[WIP] build: Quality on GH Actions (#28561)

mariajgrimaldi:

PR #554, opened 2021-10-06
refactor: replace CircleCI config with noop job

mariajgrimaldi:
**Description**
Replace current obsolete configuration so other jobs won't fail.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,57 +1,12 @@
-machine:
-  python:
-    version: 2.7.10
+version: 2.1
 
-general:
-  artifacts:
-    - "reports"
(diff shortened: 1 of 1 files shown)


felipemontoya:
@mariajgrimaldi perfect. I think this works for what we want.

I'm thinking we should merge all the prs #549 #553 #550 #554 and the one for removing the upstream github workflows in just one PR with everything squashed and call it FFI-8. Would you approve that approach? or should we merge independent PRs?

mariajgrimaldi:
I'm good with it, @felipemontoya. I'm going to open the official FFI-8 :)

PR #565, opened 2021-10-28
FFI-8: Install edunext requirements in github actions

magajh:
Part 1 of FFI-8 
This PR installs the edunext requirements in GitHub actions since they are needed for the tests to pass.
It also removes the reserved keyword checker, since it has problems with our models

### DIFF
--- a/.github/workflows/cms-tests.yml
+++ b/.github/workflows/cms-tests.yml
@@ -38,6 +38,7 @@ jobs:
         run: |
           sudo pip install --ignore-installed -r requirements/pip.txt
           sudo pip install --ignore-installed -r requirements/edx/testing.txt
+          sudo pip install --exists-action w -r requirements/edunext/base.txt
           sudo pip install "django~=${{ matrix.django-version }}.24"
       
       - name: Run tests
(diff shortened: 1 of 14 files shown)


mariajgrimaldi:
Let's comment on [this](https://github.com/eduNEXT/edx-platform/blob/7207373a77acd4d0440336fcc90242b63b3a9ddc/scripts/generic-ci-tests.sh#L137) line. It checks:
```
    """
    Check Django model fields for names that conflict with a list of reserved keywords.
    """
```
We also use the `key` keyword in other models, so this error will be raised more often than not. Also, I can't find a good reason to keep that linter.

magajh:
> Let's remove [this](https://github.com/eduNEXT/edunext-platform/blob/master/scripts/generic-ci-tests.sh#L139) line. It checks:
> 
> ```
>     """
>     Check Django model fields for names that conflict with a list of reserved keywords.
>     """
> ```
> 
> We also use the `key` keyword in other models, so this error will be raised more often than not. Also, we're not going to change a model definition to please a linter.

@mariajgrimaldi I agree. Thanks for the suggestion :)

mariajgrimaldi (approved):

PR #569, opened 2021-10-28
revert: HA-10

mariajgrimaldi:
**Description**
Feature deprecated in favor of https://github.com/edx/edx-platform/pull/15133


black-redoc:
> That PR (#15133) was merged in 2017, then why HA-10 was being applied in Hawthorn, Ironwood and Juniper? The language selector is limited to the header in Studio (in the LMS can be at either the header or the footer), but I don't think that's reason enough. The PR in hawthorn #240 mentions a problem with setting the cookie, was that solved?

Our feature isn't consistent with the language when the page is reloaded o redirected. So we should need todo some extra steps for maintain that. For that reason we need to deprecate our feature and continue with the openedx feature to manage the language selection in studio

black-redoc (approved):

PR #577, opened 2021-11-08
fix: fixed lang selector style bugs

black-redoc:
**background:**
This change is a fix for [issue: ha-10](https://3.basecamp.com/3966315/buckets/8050706/todos/4274574801#__recording_4322386730). The styles in firefox appear broken into the header for any reason that I disown. I propose this changes to resolve this. If you think in another way to fix this you are welcome to advise anything.

**Before:**
error

**After:**
Screenshot from 2021-11-08 16-00-02

### DIFF
--- a/cms/static/sass/elements/_header.scss
+++ b/cms/static/sass/elements/_header.scss
@@ -1,5 +1,19 @@
 // studio - elements - global header
 // ====================
+/*
+ * This style only works on firefox, because currently the lang selector
+ * is producing an overlapping into the header
+ */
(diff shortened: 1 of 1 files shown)


mariajgrimaldi:
Can you add a before and after so we can visualize these changes better?

black-redoc:
> Can you add a before and after so we can visualize these changes better?


this is done

mariajgrimaldi (approved):

MaferMazu (approved):

PR #585, opened 2021-11-23
JU-3: CertificateGenerated.save proxy

mariajgrimaldi:
**Description**
This PR allows sending CertificateGenerated signals with the appropriate sender, so the signal receivers work correctly.

**How to test**
_TEST 1: receivers are working_
1. Add a receiver in certificates/models.py
```
@receiver(COURSE_CERT_AWARDED, sender=GeneratedCertificate)
def received_cert_awarded_signal(sender, user, course_key, **kwargs):
    log.info("SIGNAL RECEIVED FROM: %s", type(sender))
```
2. Create a GeneratedCertificate using Django admin
3. Expected result:
`2021-11-23 19:26:58,171 INFO 596 [lms.djangoapps.certificates.models] [user 3] [ip 172.22.0.1] models.py:663 - SIGNAL RECEIVED FROM: <class 'django.db.models.base.ModelBase'>`

_TEST 2: prerequisite courses are working_
1. Configure course prerequisite
2. Complete the prerequisite course, check the user's milestones
3. Expected result: the student can access the course with the prerequisite

**Supporting information**
[Docs](https://docs.google.com/document/d/1BtEShTyy5I9g_noIX0vA70dp_Oq4dCqKjCUdQN9xM5o/edit?usp=sharing)

### DIFF
--- a/lms/djangoapps/certificates/models.py
+++ b/lms/djangoapps/certificates/models.py
@@ -443,8 +443,12 @@ class GeneratedCertificate(models.Model):
         As well as the COURSE_CERT_CHANGED for any save event.
         """
         super().save(*args, **kwargs)
+        if self._meta.proxy:
+            sending_class = self._meta.proxy_for_model
+        else:
+            sending_class = self.__class__
         COURSE_CERT_CHANGED.send_robust(
-            sender=self.__class__,
+            sender=sending_class,
             user=self.user,
             course_key=self.course_id,
             mode=self.mode,
@@ -475,7 +479,7 @@ class GeneratedCertificate(models.Model):
 
         if CertificateStatuses.is_passing_status(self.status):
             COURSE_CERT_AWARDED.send_robust(
-                sender=self.__class__,
+                sender=sending_class,
                 user=self.user,
                 course_key=self.course_id,
                 mode=self.mode,


magajh (approved):
lgtm

lms/djangoapps/certificates/models.py
```diff
@@ -442,9 +442,14 @@ def save(self, *args, **kwargs):  # pylint: disable=signature-differs
         signal iff we are saving a record of a learner passing the course.
         As well as the COURSE_CERT_CHANGED for any save event.
         """
-        super().save(*args, **kwargs)
+        if GeneratedCertificate.__base__.__name__ == 'GeneratedCertificate':
+            super(GeneratedCertificate.__base__, self).save(*args, **kwargs)
```
MoisesGSalas, line 446: Why are we doing this? Doesn't the proxy already use the correct save? https://github.com/eduNEXT/eox-tenant/blob/0795d3cfc2161f5c1513af74d5e47b071fe84647/eox_tenant/tenant_wise/proxies.py#L268

lms/djangoapps/certificates/models.py
```diff
@@ -442,9 +442,14 @@ def save(self, *args, **kwargs):  # pylint: disable=signature-differs
         signal iff we are saving a record of a learner passing the course.
         As well as the COURSE_CERT_CHANGED for any save event.
         """
-        super().save(*args, **kwargs)
+        if GeneratedCertificate.__base__.__name__ == 'GeneratedCertificate':
```
MoisesGSalas, line 445: This is to check if you are in a proxy? maybe you can use `self._meta.proxy`

lms/djangoapps/certificates/models.py
```diff
@@ -442,9 +442,14 @@ def save(self, *args, **kwargs):  # pylint: disable=signature-differs
...
+            super(GeneratedCertificate.__base__, self).save(*args, **kwargs)
+            sending_class = GeneratedCertificate.__base__
```
MoisesGSalas, line 447: And this could be

```suggestion
            sending_class = self._meta.proxy_for_model
```

lms/djangoapps/certificates/models.py
mariajgrimaldi: is it safer?

MoisesGSalas: It seems like a more appropriate check instead of comparing names. If you're not behind a proxy it will return false.

lms/djangoapps/certificates/models.py
mariajgrimaldi: I don't think that's gonna work :(
image

MoisesGSalas: You should check `self._meta.proxy` not `GeneratedCertificate._meta.proxy`

mariajgrimaldi: This happened when I did:
image

MoisesGSalas: But you shouldn't care about `GeneratedCertificate`, the real value should be the one in self. In the end what we want to know is if the current object (aka `self`) is a proxy or not. If it is, then use the parent model class as the sender, if it isn't then use it's own class.

MoisesGSalas: The problem I'm seeing is that the behavior of `GeneratedCertificate` is the oposite of `self`. When one is a proxy the other isn't.

In that screenshot you are doing `if GeneratedCertificate._meta.proxy`, if you do `self._meta.proxy` it would make more sense.

Now, why is it behaving like this I don't know. eox-tenant supposedly [monkey patchs the model after app registry](https://github.com/eduNEXT/eox-tenant/blob/master/eox_tenant/apps.py#L100). Maybe is some interaction with the Django admin that I don't know.

lms/djangoapps/certificates/models.py
```diff
@@ -443,10 +443,10 @@ def save(self, *args, **kwargs):  # pylint: disable=signature-differs
         As well as the COURSE_CERT_CHANGED for any save event.
         """
         super().save(*args, **kwargs)
-        if GeneratedCertificate._meta.proxy:
-            sending_class = GeneratedCertificate.__base__
+        if self._meta.proxy:
+            sending_class = self._meta.proxy_for_model
         else:
-            sending_class = GeneratedCertificate
+            sending_class = GeneratedCertificate.__base__
```
MoisesGSalas, line 449:
```suggestion
            sending_class = self.__class__
```
Now that we are sure `self` is not a proxy the sender can be the same as it was originally.

https://github.com/eduNEXT/edunext-platform/blob/2803f5faff91ccbdc466f0d1a904b9b8a33af881/lms/djangoapps/certificates/models.py#L400

MoisesGSalas (approved):

PR #590, opened 2021-11-26
LI-16: Renable the old IDV views

MoisesGSalas:
## Description
Use the Django views for the Identity Verification Service instead of relying on the account MFE.

This is essentially a revert of: 
- 00ad36839d73e47a8f82dc5272222c150bda5e4e 
- 4d19d9687cd4366735967bd072fe077f2dc20969
- 2a4c26c3675076657806e5f228669bc5522f78cd

Previously performing the IDV via the account MFE was behind a waffle flag that was removed, I opted to use a new setting `ENABLE_ACCOUNT_MFE` instead of the original `redirect_to_idv_microfrontend` (which relies on the flag).

## Testing instructions
There are several ways to access the verification views, most notably the 'Verify Now' button that appears in the dashboard listing once you enroll in a course in a verified mode.

1. Create course with a verified mode.
2. Enroll in the course, make sure your enrollment type is the verified mode you chose (verified, professional)
3. Click on the 'Verify Now' link and perform the verification flow.

## Additional Info
The usual way to create a verified course is through the ecommerce service. Currently we don't have access to the ecommerce service, neither on stage or our tutor based development tools, so we have to take some manual steps.

1. To create course with a verified track, we create a new course and add two entries to the course modes table for said course: one for audit and one for verified.
2. We comment[ this call](https://github.com/eduNEXT/edunext-platform/blob/edunext/limonero.master/lms/djangoapps/verify_student/views.py#L398) to the ecommerce service and set `processors=""`

### DIFF
--- a/common/djangoapps/course_modes/views.py
+++ b/common/djangoapps/course_modes/views.py
@@ -111,7 +111,8 @@ class ChooseModeView(View):
         has_enrolled_professional = (CourseMode.is_professional_slug(enrollment_mode) and is_active)
         if CourseMode.has_professional_mode(modes) and not has_enrolled_professional:
             purchase_workflow = request.GET.get("purchase_workflow", "single")
-            redirect_url = IDVerificationService.get_verify_location(course_id=course_key)
+            verify_url = IDVerificationService.get_verify_location('verify_student_start_flow', course_id=course_key)
+            redirect_url = f"{verify_url}?purchase_workflow={purchase_workflow}"
             if ecommerce_service.is_enabled(request.user):
                 professional_mode = modes.get(CourseMode.NO_ID_PROFESSIONAL_MODE) or modes.get(CourseMode.PROFESSIONAL)
                 if purchase_workflow == "single" and professional_mode.sku:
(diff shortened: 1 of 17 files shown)


common/djangoapps/course_modes/views.py
```diff
@@ -111,7 +111,8 @@ def get(self, request, course_id, error=None):  # lint-amnesty, pylint: disable=
         has_enrolled_professional = (CourseMode.is_professional_slug(enrollment_mode) and is_active)
         if CourseMode.has_professional_mode(modes) and not has_enrolled_professional:
             purchase_workflow = request.GET.get("purchase_workflow", "single")
-            redirect_url = IDVerificationService.get_verify_location(course_id=course_key)
+            verify_url = IDVerificationService.get_verify_location('verify_student_start_flow', course_id=course_key)
+            redirect_url = "{url}?purchase_workflow={workflow}".format(url=verify_url, workflow=purchase_workflow)
```
mariajgrimaldi, line 115: what about f strings?

mariajgrimaldi (approved):

PR #591, opened 2021-11-27
feat: Add token which disables different origin subframe

MaferMazu:
## Description

This PR is part of the changes that were made after the migration cut, but it is necessary for bragi to work in limonero.

- [Cherry-pick](https://github.com/eduNEXT/edx-platform/commit/d6077850a88eb3d823b0f628146c4e7ca0a6a246)

Issues that need this change to work in limonero:

- [JU-11 Bragi](https://edunext.atlassian.net/browse/PS2021-1266?atlOrigin=eyJpIjoiYTRlMThiMjJmYTMxNGI3OGI1YjI0Nzc5ZTVlYTg1MWYiLCJwIjoiaiJ9)
- [Bragi](https://edunext.atlassian.net/browse/PS2021-1264?atlOrigin=eyJpIjoiYmYzMWY1ZWUzY2U1NGFmNWI2MDY4YWVjMjRhOWQ2ZTgiLCJwIjoiaiJ9)

Without this change the bragi `main.html` causes this error to appear:
Screenshot from 2021-11-26 16-33-39

### DIFF
--- a/lms/envs/common.py
+++ b/lms/envs/common.py
@@ -4671,3 +4671,7 @@ LOGO_URL_PNG = None
 LOGO_TRADEMARK_URL = None
 FAVICON_URL = None
 DEFAULT_EMAIL_LOGO_URL = 'https://edx-cdn.org/v3/default/logo.png'
+
+################# Settings for Chrome-specific origin trials ########
+# Token for " Disable Different Origin Subframe Dialog Suppression" for http://localhost:18000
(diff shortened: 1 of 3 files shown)


MoisesGSalas:
I'm concerned about this part from the original PR:

> This fix is only a temporary workaround for Chrome browsers! It will only work until around mid-December. A more permanent solution needs to be designed and implemented well before that time.

I read bit of the rationale in the PR and it seems like this only applies to the learning MFE?
Given that we aren't going to use it soon I would rather remove the setting from the template. 

>  Chrome 92 suppressed different origin subframe dialog/modal creation via JS, which broke the "Open in New Window" mode of the LTI Component unit as well as ORA unit modals for the Learning MFE. But Google provided an `origin trial` which can disable this suppression. A change was added to allow a Lilac Open edx installation using the Learning MFE to specify a token which enables the Chrome origin trial which disables the suppression. To do so, generate a token on this `registration page` and provide it as a custom Django setting named: ``CHROME_DISABLE_SUBFRAME_DIALOG_SUPPRESSION_TOKEN``.

MaferMazu:
Thanks @MoisesGSalas i am going to tell this to @jignaciopm to see if we remove that variable from the template.

PR #593, opened 2021-12-03
feat: Codejail as external service (optional)

ericfab179:
Backport of changes merged in this PR https://github.com/edx/edx-platform/pull/27795/files 

Add codejail service settings (endpoint and feature flag).
Add conditional to allow running codejail using a REST API service when flag is enabled.

### DIFF
--- a/cms/envs/common.py
+++ b/cms/envs/common.py
@@ -973,6 +973,28 @@ CODE_JAIL = {
 
 COURSES_WITH_UNSAFE_CODE = []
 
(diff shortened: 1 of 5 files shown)


Alec4r:
#Why do I merge it without approval?
I talked with Felipe about this and this has been tested enough, so there isn't a reason to don't merge it.

felipemontoya:
I'm late here but just for the record. :+1:

PR #598, opened 2021-12-14
Li/ednx/li6

black-redoc:
**Description:**
Add more description to JWT requests.

**Commits:**
- 1a73b271
- 0c7ac618

**Jira:**
- [issue](https://edunext.atlassian.net/browse/PS2021-1185)

**Notes:**
To test this result is necesary set the `"JWT_EXTENDED_PROFILE": true` into tenant configs. And put `{"session_id": "l0m9wx1bfvptcew69skaazmyye4vigo4", "personal_id":"12345"}` into user's profile meta settings. You can get the `<client-secret>` and `<client-id>` in OAuth-Toolkit/Applications/<Your-Tenant> in the django admin. Then the `access_token` into the response,  you can use https://jwt.io to decode it.

**Tests:**
```bash
curl -i -X POST http://tenant-a.lms.limonero.edunext.link:8000/oauth2/access_token/ \
-F "client_id=<client-id>" \
-F "client_secret=<client-secret>" \
-F "grant_type=client_credentials" \
-F "token_type=jwt"
```

**Expected result:**
Screenshot from 2021-12-14 17-47-33

### DIFF
--- a/openedx/core/djangoapps/oauth_dispatch/jwt.py
+++ b/openedx/core/djangoapps/oauth_dispatch/jwt.py
@@ -11,6 +11,7 @@ from jwkest import jwk
 from jwkest.jws import JWS
 
 from common.djangoapps.student.models import UserProfile, anonymous_id_for_user
+from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
 
 
 def create_jwt_for_user(user, secret=None, aud=None, additional_claims=None, scopes=None):
@@ -178,9 +179,23 @@ def _attach_email_claim(payload, user):
 
 def _attach_profile_claim(payload, user):
     """Add the profile claim details to the JWT payload."""
+    extended_profile_fields = None
+
     try:
         # Some users (e.g., service users) may not have user profiles.
-        name = UserProfile.objects.get(user=user).name
+        profile = UserProfile.objects.get(user=user)
+        name = profile.name
+
+        if profile.meta and configuration_helpers.get_value('JWT_EXTENDED_PROFILE', False):
+            try:
+                extended_profile_fields = {
+                    key: value
+                    for key, value in json.loads(profile.meta).items()
+                    if key in configuration_helpers.get_value('extended_profile_fields', [])
+                }
+            except ValueError:
+                pass
+
     except UserProfile.DoesNotExist:
         name = None
 
@@ -190,6 +205,7 @@ def _attach_profile_claim(payload, user):
         'given_name': user.first_name,
         'administrator': user.is_staff,
         'superuser': user.is_superuser,
+        'extended_profile_fields': extended_profile_fields,
     })
 
 


MaferMazu:
Could you please add how to test it to get a result similar to yours? Could you add it to the documentation as well?

black-redoc:
> Could you please add how to test it to get a result similar to yours? Could you add it to the documentation as well?

@MaferMazu this is done!

openedx/core/djangoapps/oauth_dispatch/jwt.py
```diff
@@ -178,9 +180,27 @@ def _attach_email_claim(payload, user):
...
+                    key: value
+                    for key, value in json.loads(profile.meta).items()
+                    if key in configuration_helpers.get_value('extended_profile_fields', [])
+                }
+            except ValueError:
+                pass
+        fields = configuration_helpers.get_value('extended_profile_fields', [])
+
+        if fields and configuration_helpers.get_value('JWT_EXTENDED_PROFILE', False):
+            extended_profile_fields = {key: value for key, value in json.loads(profile.meta).items() if key in fields}
```
MaferMazu, line 202: You have to be careful because here you added both changes of the commits but in the last one the last part is no longer within the code

```python
fields = configuration_helpers.get_value('extended_profile_fields', [])

if fields and configuration_helpers.get_value('JWT_EXTENDED_PROFILE', False):
     extended_profile_fields = {key: value for key, value in json.loads(profile.meta).items() if key in fields}
```

here you can see what was left: https://github.com/eduNEXT/edunext-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/jwt.py#L179

Try removing that piece of code (the last part) to see if it works as it should.

openedx/core/djangoapps/oauth_dispatch/jwt.py
```diff
@@ -11,6 +11,8 @@
 from jwkest.jws import JWS
 
 from common.djangoapps.student.models import UserProfile, anonymous_id_for_user
+from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
+from common.djangoapps.student.models import UserProfile, anonymous_id_for_user
```
MaferMazu, line 15: You have two equal imports

openedx/core/djangoapps/oauth_dispatch/jwt.py
black-redoc: This is done!

openedx/core/djangoapps/oauth_dispatch/jwt.py
black-redoc: this is done!

openedx/core/djangoapps/oauth_dispatch/jwt.py
```diff
@@ -178,9 +179,24 @@ def _attach_email_claim(payload, user):
...
+        if profile.meta and configuration_helpers.get_value('JWT_EXTENDED_PROFILE', False):
+            try:
+                extended_profile_fields = {
+                    key: value
+                    for key, value in json.loads(profile.meta).items()
+                    if key in configuration_helpers.get_value('extended_profile_fields', [])
+                }
+            except ValueError:
+                pass
+        fields = configuration_helpers.get_value('extended_profile_fields', [])
```
MaferMazu, line 198: I think this variable is unnecessary
https://github.com/eduNEXT/edunext-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/jwt.py#L179

black-redoc: this is done!

MaferMazu (approved):

mariajgrimaldi:
Please, don't merge this until you test it on a stage env.

MoisesGSalas (approved):

PR #602, opened 2021-12-17
Fix to prevent 500 error when course mode currency is not the same as…

black-redoc:
**Background:**
This PR solves the http 500 error because of the `CURRENCY PAID MODE` is setted different `USD` the currency default in the django settings and the course created is another one.

**Commits:**
- 38959ca
**Jira:**
- https://edunext.atlassian.net/browse/PS2021-1165

**After:**
Screenshot from 2021-12-17 10-52-37

**Before:**
Screenshot from 2021-12-17 10-54-43

### DIFF
--- a/common/djangoapps/course_modes/models.py
+++ b/common/djangoapps/course_modes/models.py
@@ -762,7 +762,10 @@ class CourseMode(models.Model):
         If there is no mode found, will return the price of DEFAULT_MODE, which is 0
         """
         modes = cls.modes_for_course(course_id)
-        return min(mode.min_price for mode in modes if mode.currency.lower() == currency.lower())
+        return min(
+            (mode.min_price for mode in modes if mode.currency.lower() == currency.lower()),
+            default=0
+        )
 
     @classmethod
     def is_eligible_for_certificate(cls, mode_slug, status=None):


common/djangoapps/course_modes/models.py
```diff
@@ -762,7 +762,10 @@ def min_course_price_for_currency(cls, course_id, currency):
         If there is no mode found, will return the price of DEFAULT_MODE, which is 0
         """
         modes = cls.modes_for_course(course_id)
-        return min(mode.min_price for mode in modes if mode.currency.lower() == currency.lower())
+        try:
+            return min(mode.min_price for mode in modes if mode.currency.lower() == currency.lower())
+        except ValueError:
+            return 0
```
MoisesGSalas, line 768:
```suggestion
        return min(
            (mode.min_price for mode in modes if mode.currency.lower() == currency.lower()),
            default=0,
        )

```

black-redoc: this is done!

MoisesGSalas: Not quite, the idea is to replace the whole `try-except` by using a default value for `min`. Also the tests are failing because you must surround `mode.min_price for mode in modes if mode.currency.lower() == currency.lower()` in parenthesis.

common/djangoapps/course_modes/models.py
```diff
@@ -762,7 +762,13 @@ def min_course_price_for_currency(cls, course_id, currency):
         If there is no mode found, will return the price of DEFAULT_MODE, which is 0
         """
         modes = cls.modes_for_course(course_id)
-        return min(mode.min_price for mode in modes if mode.currency.lower() == currency.lower())
+        try:
```
MoisesGSalas, line 765: no need for the try except

black-redoc: this is done!

MoisesGSalas (approved):

PR #606, opened 2021-12-23
FFI-8 P3: Fix python package installation

MoisesGSalas:
## Description
- Ensure the correct dependencies are installed even with a broken version in the cache.
- Move from `continue-on-error: true` to `fail-fast: false` to avoid green checks even when all jobs failed.
- Install eduNEXT dependencies on the pylint workflows. There was an import error due to eox-tenant that broke the pylint run. Because pylint is run via paver the exit code is 0 when an error is found and the result of the job was a success. This pylint violations seem to have been silenced for a few weeks now.

### DIFF
--- a/.github/workflows/cms-tests.yml
+++ b/.github/workflows/cms-tests.yml
@@ -36,10 +36,10 @@ jobs:
 
       - name: Install requirements
         run: |
-          sudo pip install --ignore-installed -r requirements/pip.txt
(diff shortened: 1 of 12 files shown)


JuanDavidBuitrago (approved):

johanv26 (approved):
Good solution, this way we discover some pylint issues we have passed by the paver in the tests.
So in the future, we can change them.

PR #614, opened 2022-01-12
translations: downloaded lilac translations

black-redoc:
**description:**
Updated translations and downloaded from transifex.

**changes:**
updated i18n and static files from:
- lms/static
- cms/static
updated locales from:
- conf/locale

**how to test:**
check the pages and change the translations with the language selector into header.

**jira:**
https://edunext.atlassian.net/browse/PS2021-1374


.tx/config
```diff
@@ -78,3 +78,4 @@ file_filter = conf/locale/<lang>/LC_MESSAGES/djangojs.po
 source_file = conf/locale/en/LC_MESSAGES/djangojs.po
 source_lang = en
 type = PO
+
```
mariajgrimaldi, line 81: is this necessary?

MoisesGSalas: No, this file should stay the same.

MoisesGSalas (changes requested):
There is a lot of files missing. Filtering by `*.po` extension only throws 2 files: `conf/locale/th/LC_MESSAGES/django.po` and `conf/locale/th/LC_MESSAGES/djangojs.po`

black-redoc:
> There is a lot of files missing. Filtering by `*.po` extension only throws 2 files: `conf/locale/th/LC_MESSAGES/django.po` and `conf/locale/th/LC_MESSAGES/djangojs.po`

What do you mean?

mariajgrimaldi (approved):

PR #624, opened 2022-01-20
fix: fix i18n translations chore

MoisesGSalas:
## Description

Some translations weren't validated and generated some errors while rendering some pages (a page that should throw a 404 ended with a 500 instead). Some fixes were to allow `i18n_tool validate` pass without errors.


johanv26:
It works in stage lilac.
**Before**:


**After**:

johanv26 (approved):

mariajgrimaldi (approved):

JuanDavidBuitrago (approved):

PR #630, opened 2022-02-17
refactor: use Django ORM instead of modulestore when getting course summaries

mariajgrimaldi:
## Description

This PR improves studio course listing performance for users with course access roles like staff or instructor, by changing from MongoDB queries to Mysql queries.

We measured the improvements as follows:

After 100 requests:

| Number of courses | Avg time elapsed before (s)  | Avg time elapsed after (s) |
|---------------------------|---------------------------------------|------------------------------------|
|         50                   |                2.34                        |         2.28                           |
|         107                 |                3.16                        |         2.57                           |
|         548                 |                7.91                        |         3.16                           |
|         1069               |                Error 500                |         4.62                           |
|         3380               |                Error 500                |         8.21                           |

If you go to /home using users with access roles staff/instructor in an org with more than ~500 courses, the page will collapse and raise a Network Timeout.

**Versions**
Mongo 4.0
MySQL 5.7
Lilac release

## Testing instructions

Go to Studio Home with a user with access roles (instructor/staff) over an organization, then the list of courses must match the CourseOverviews.

## Other information

According to [this](https://discuss.openedx.org/t/adr-for-removing-mongodb-from-edx-platform/6001), this change won't be necessary after deprecating MongoDB from edxapp.

I cherry-picked from [upstream PR](https://github.com/openedx/edx-platform/pull/29944). I'm not sure whether they're going to accept the PR, but the main goal is to discuss our performance issues and this solution.

### DIFF
--- a/cms/djangoapps/contentstore/tests/test_course_listing.py
+++ b/cms/djangoapps/contentstore/tests/test_course_listing.py
@@ -33,6 +33,8 @@ from common.djangoapps.student.roles import (
 )
 from common.djangoapps.student.tests.factories import UserFactory
 from openedx.core.djangoapps.content.course_overviews.tests.factories import CourseOverviewFactory
+from openedx.core.djangoapps.content.course_overviews.models import CourseOverview
+from openedx.core.djangoapps.waffle_utils.testutils import WAFFLE_TABLES
 from xmodule.course_module import CourseSummary
 from xmodule.modulestore import ModuleStoreEnum
 from xmodule.modulestore.tests.django_utils import ModuleStoreTestCase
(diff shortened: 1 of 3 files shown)


Ian2012 (approved):

jfavellar90:
@mariajgrimaldi I'm curious about the way you found out the bottleneck was that particular Mongo query

mariajgrimaldi:
> @mariajgrimaldi I'm curious about the way you found out the bottleneck was that particular Mongo query

[Here](https://sentry.io/organizations/edunext/issues/2944641018/?project=6134348&query=is%3Aunresolved) is the sentry error caught when replicating the issue, there, there's an interesting method call: 
`xmodule/modulestore/split_mongo/split.py in get_course_summaries at line 1055`
get_course_summaries is called in two places (not tests) in courses.py: in [_accessible_courses_summary_iter](https://github.com/eduNEXT/edunext-platform/blob/master/cms/djangoapps/contentstore/views/course.py#L412) and [_accessible_courses_list_from_groups](https://github.com/eduNEXT/edunext-platform/blob/master/cms/djangoapps/contentstore/views/course.py#L506)
From what I gathered, the first method is called in two situations:
1. Global Staff users
2. When users had old groups
And the second is with users with course access roles like staff/instructor (our exact problem). Besides, why does it work for global staff users with studio optimization? Maybe because it also uses the Django ORM instead of modulestore.

mariajgrimaldi:
I'll be fixing those tests soon

felipemontoya (approved):
Given the relative approval at the larger communiy, I think we can merge this safely once tests pass.

+1

cms/djangoapps/contentstore/tests/test_course_listing.py
```diff
@@ -281,17 +290,6 @@ def test_course_listing_performance(self, store, courses_list_from_group_calls,
...
-        with check_mongo_calls(courses_list_from_group_calls):
-            _accessible_courses_list_from_groups(self.request)
-
-        with check_mongo_calls(courses_list_calls):
-            list(_accessible_courses_iter_for_tests(self.request))
-        # Calls:
-        #    1) query old mongo
-        #    2) get_more on old mongo
-        #    3) query split (but no courses so no fetching of data)
-
```
mariajgrimaldi, line 294: for mysql

felipemontoya:
This passed nicely. @DonatoB-eduNEXT are we ready to merge and deploy this?

DonatoB-eduNEXT:
Of course! This is already going to prod

PR #632, opened 2022-02-21
fix: icon on course outline was not showing correctly

Ian2012:
## Description

On the course outline, the icon for graded and scored problems was showing inconsistently. The cause was the parameter of the any() function. Before the change it was a generator so, the function recurse_mark_scored(child) was not called for all children, now it receives a list comprehension with the result of recurse_mark_scored(child) for all children.

When the generator parameter was passed, only the first block that meets the condition was marked as scored. With this change now all blocks that meet the condition were marked as scored.

## Testing instructions

1. On Studio go to a course
2. Mark some subsections as graded (configure a grading policy is necessary)
3. Add a problem to the subsections you marked as graded.
4. On LMS you should see the icon for every graded and scored units.

Before:
image
After:
image


MoisesGSalas (approved):

DonatoB-eduNEXT (approved):

mariajgrimaldi (approved):

mariajgrimaldi:
Excellent job! can we take this upstream? :partying_face:

PR #635, opened 2022-04-14
ci: add distro nightly tag to github workflows

Alec4r:
## Description
This commit is a proposal to create each day one tag with the changes of the day.

This may be the first part of the CI that will build a new image for each nightly tag.

## How it works
This has a cronjob that will run daily at 04:00 (UTC) or 23:00 (Colombia TIme), then, this will check if there are new commits in a range of 23 hours if there aren't this won't create a tag.

### Tag format
```
nightly/vL.N20200414d2d0a67
         | |   |    |___ Last commit's hash (d2d0a67)
         | |   |___ Current date (2020-04-14)
         | |___ N: Nightly tag
         |___ eduNEXT distro release
```

### DIFF
--- a/.github/workflows/distro-nightly-tag.yml
+++ b/.github/workflows/distro-nightly-tag.yml
@@ -1,0 +1,27 @@
+name: Distro nigthly tag
+
+on:
+  schedule:
+    - cron: '00 04 * * *'
+
+jobs:
+  nightly-tag:
+    runs-on: ubuntu-20.04
+    name: Create nightly tag
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+      - name: get the last commit of day
+        id: last-commit-of-day
+        run: |
+          hash=$(git --no-pager log -1 --format="%h" --since=23.hours)
+          echo "::set-output name=COMMITHASH::${hash}"
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=NOW::$(date +'%Y%m%d')"
+      - name: Tag commit
+        uses: tvdias/github-tagger@v0.0.2
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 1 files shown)


Alec4r:
It was tested on a personal fork: https://github.com/Alec4r/edunext-platform/runs/6029454880?check_suite_focus=true

felipemontoya:
Would this tag the release every night? or only when there are changes?

felipemontoya:
I think this does exactly what is advertised.

Can you make it so, that the tag is `nightly/vL.N20200414d2d0a67` so that on most GUI clients, all the nightly tags can be collapsed as if they were in a directory?

Do you think we would like to have more than one nightly tag? for example, one for every open-release we currently support?
Or maybe I got it wrong and this is only for master and we will use it mostly for testing the integration with plugins and whatnot.

Anyways if you want to have this merged, go ahead, I'm good.

felipemontoya (changes requested):
Just to mention again the tag in a `nightly/` namespace,

Alec4r:
> I think this does exactly what is advertised.
> 
> Can you make it so, that the tag is `nightly/vL.N20200414d2d0a67` so that on most GUI clients, all the nightly tags can be collapsed as if they were in a directory?
> 
> Do you think we would like to have more than one nightly tag? for example, one for every open-release we currently support? Or maybe I got it wrong and this is only for master and we will use it mostly for testing the integration with plugins and whatnot.
> 
> Anyways if you want to have this merged, go ahead, I'm good.

1. Yes, I will change the tag.
2.  We can have more than one, but for that, we must set each one in the action, or maybe we can use GitHub environment variables.
3. We may begin just with the master branch, It's enough for me.

felipemontoya (approved):

MoisesGSalas (approved):
I'm fine doing it like this, but I wonder if it wouldn't be better to simply retag de newest commit each day and have only one single nightly tag.

MaferMazu:
I talked with @Alec4r, who told me that this is not on our radar and that we can reopen it if necessary. I will close this.

PR #645, opened 2022-04-22
N11 - fix: correct relative dates flag

MaferMazu:
## Description

Correct relative dates flag and use CourseWaffleFlag

## Other info

Juniper commit:
https://github.com/fccn/edx-platform/commit/84412e2fb235f9a9ec33009713a048f6f4ebed21

[Jira issue](https://github.com/fccn/edx-platform/commit/62f1682ea0171b7b4d5e1fb6614446bbf7d09c3e)
[Migration excel](https://docs.google.com/spreadsheets/d/1QCF3H_rYWm4wRCcchCV0UqhQjjKYUQOKneywkU2IM88/edit?usp=sharing)


Henrrypg:
About the adjustements i made.

some of these was module imports, add missing files and fix tests. WafflerFlags is being used a different structure than ours. but essentially does the same thing (They basically made his class where they used methods like overwrite and we use the overwrite as decorator). I just added the missing overwrite in our structure with their logic and change it in the tests.

PR #668, opened 2022-09-27
backport grant course creation permission

Ian2012:
This PR backports the functionality of managing course creator permissions from the CourseCreator table which is still kind of compatible with our custom development performed in [BC-14](https://github.com/eduNEXT/edunext-platform/pull/574/files) but adds another layer of verification as these changes are only in the backend.

Regarding the frontend there is a PR on upstream to add the selector we currently use.

For this to work, we still need to revert all commits related to our custom development and only let this one. Also, we need to migrate the permissions for courseaccessrole to CourseCreator

Upstream [PR](https://github.com/openedx/edx-platform/pull/26616)


felipemontoya:
for context: this work should be merged before the nutmeg migration, but we need to ensure that we do every step in a stable manner.

Probably something like this:

1) only BC-14 way (current status)
2) we add this while keeping the UI working both with BC-14 and this.
3) we make eox-manage use this backend instead of BC-14
4) we migrate all old data from BC-14 into this
5) we remove BC-14 code and data

DonatoBD (approved):

PR #683, opened 2022-10-25
feat: Add setting DISABLE_ORDER_HISTORY_TAB

andrey-canon:
## Description

This will add a setting that allows to hide the Order history tab  when the order history microfrontend is not available 

## Testing instructions

Add in settings  `DISABLE_ORDER_HISTORY_TAB = True`  to hide
**This only work when eox-tenant is active** 

**Before**
image

**After**
image


johanv26 (approved):
Nice OR, this also doesn't modify the method `should_redirect_to_order_history_microfrontend`
And also with default false value.

PR #684, opened 2022-10-26
feat: adding persistent grade event

DeimerM:
## Persistent Grade Event

Added new event to the platform, this event occurs when a persistent grade is created or updated, for further information about how to test this event you can see the following PR https://github.com/openedx/edx-platform/pull/30916


mariajgrimaldi:
> @mariajgrimaldi I'm not sure if we can skip that many versions of openedx-events. What do you think?

We can, but before, check for breaking changes in the openedx-events changelog and ensure they don't break anything here.

PR #701, opened 2022-11-23
feat: Add email field to UserMappingView

andrey-canon:
## Description

This will add the email field in the API response, this is necessary since the client needs to know the email value in order to use it in another process 

## Note
This is required to test subscription workflow in stage, and depends on the client's decision to move to production 

## Testing instructions

- Get jwt token in /oauth2/access_token/
- Set Authorization header with the previous token and make a request to  /api/third_party_auth/v0/providers/< provider-id >/users provider-id is the slug plus the backend prefix, for example   saml-default-slug the slug can be found in /admin/third_party_auth/samlproviderconfig/

### DIFF
--- a/common/djangoapps/third_party_auth/api/serializers.py
+++ b/common/djangoapps/third_party_auth/api/serializers.py
@@ -9,6 +9,7 @@ class UserMappingSerializer(serializers.Serializer):  # pylint: disable=abstract
     provider = None
     username = serializers.SerializerMethodField()
     remote_id = serializers.SerializerMethodField()
+    email = serializers.SerializerMethodField()
 
     def __init__(self, *args, **kwargs):
         self.provider = kwargs['context'].get('provider', None)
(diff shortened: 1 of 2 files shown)


johanv26:
@andrey-canon I am trying to test, but I thing I am using wrong the endpoint?
image

andrey-canon:
@johanv26 I updated the comment since the word between < > was hidden this is the url `/api/third_party_auth/v0/providers/< provider-id >/users `  provider-id is the slug plus the backend prefix  in your test should be  saml-okta-sso  and finally that configuration is set with the site edunext.stage.nelp.gov.sa so you have to test with the same site 

so the url in your case must be `https://edunext.stage.nelp.gov.sa/api/third_party_auth/v0/providers/saml-okta-sso/users`

johanv26:
Sure is working
Does this only work for third-party users?
image

andrey-canon:
> Sure is working Does this only work for third-party users? image

yes, this is a quick implementation to get users, however I think the client will choose another implementation which allows to get all the users not just the third party

johanv26 (approved):

PR #704, opened 2022-11-25
feat: Allow to include translations for extended profile fields

andrey-canon:
## Description
These changes allow to set labels for extended profile fields in order to include translations 

## Testing instructions
1. set extended profile fields in your tenant config, example 
```
    "extended_profile_fields": [
        "hobby",
        "sport",
        "movie"
    ]
```
2. Set extended profile field translations, example
```
    "extended_profile_fields_translations": {
        "ar": {
            "hobby": "\u0647\u0648\u0627\u064a\u0629"
        }
    },
```
The `extended_profile_fields_translations` must be a dictionary with language code keys and every language code key is a dictionary that must include the field that will be translated where the key is the field name and the value is the translation

3. go to /account/settings and change the language if it is necessary 

## Results in local
just the field `hobby` was set 

EN
image

AR
image

### DIFF
--- a/openedx/core/djangoapps/user_api/accounts/settings_views.py
+++ b/openedx/core/djangoapps/user_api/accounts/settings_views.py
@@ -23,6 +23,7 @@ from openedx.core.djangoapps.dark_lang.models import DarkLangConfig
 from openedx.core.djangoapps.lang_pref.api import all_languages, released_languages
 from openedx.core.djangoapps.programs.models import ProgramsApiConfig
 from openedx.core.djangoapps.site_configuration import helpers as configuration_helpers
+from openedx.core.djangoapps.theming import helpers as theming_helpers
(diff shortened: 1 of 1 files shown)


openedx/core/djangoapps/user_api/accounts/settings_views.py
```diff
@@ -250,6 +251,13 @@ def _get_extended_profile_fields():
         "profession": _("Profession"),
         "specialty": _("Specialty")
     }
+    request = theming_helpers.get_current_request()
```
johanv26, line 254: Why do you have to use theming_helpers??

johanv26 (commented):
This translations should only work for registration form. So settings view for this PR and registration form set here: 
https://github.com/eduNEXT/eox-nelp/pull/13/files

johanv26 (approved):

PR #720, opened 2022-12-27
feat: add eox tenant compatibilty for cache programs command

johanseto:
## Description
This is related the programs functionality in the LMS and its connection to course catalog provider in our case programs configurated in discovery.

This PR adds the possibility to run `./manage.py lms cache_programs` using eox_tenant and the `course_catalog_api_url` used in the lms_config of the tenant config. (By the nelp use is the partner with form `https://{tenant_name}.discovery.stage.nelp.gov.sa/api/v1/)`
This also keeps the functionality if eox_tenant is not installed in order to continue working with site and site_configuration.

## Testing running
This was tested in stage and updated the image of a program configurated in the discovery service. 

Peek 2022-12-27 16-37

**Othe tenant update**
https://drive.google.com/file/d/1A2P7YIjl38Ps1G_UAPLqXXQpybB3Ov8X/view?usp=sharing

### DIFF
--- a/openedx/core/djangoapps/catalog/management/commands/cache_programs.py
+++ b/openedx/core/djangoapps/catalog/management/commands/cache_programs.py
@@ -26,6 +26,8 @@ from openedx.core.djangoapps.catalog.utils import (
     course_run_keys_for_program,
     course_uuids_for_program,
     create_catalog_api_client,
+    get_lms_config_and_catalog_url_by_eox_tenant,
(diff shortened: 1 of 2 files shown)


johanseto:
This PR discussion and implementation will continue with the approach of using a plugin in order to avoid modifying the code.  https://github.com/eduNEXT/eox-nelp/pull/14

PR #733, opened 2023-01-30
Jlc/mfe learning by tenant settings

johanseto:
## Description
Add the possibility to manage the learning MFE by the tenant using `USE_LEARNING_LEGACY_FRONTEND`.

**Useful information to include:**
- The main Django switches for the legacy frontend have to be turned on.

-  By default, the tenant would use legacy learning instead MFE.
- To use MFE you have to set the tenant config var `USE_LEARNING_LEGACY_FRONTEND: false`
- To load the change you have to load a different tenant each time this setting is configured.

## Usage
- changes refreshing using  `USE_LEARNING_LEGACY_FRONTEND`
Peek 2023-01-30 14-44

- changes without the `USE_LEARNING_LEGACY_FRONTEND` defined.
Peek 2023-01-30 14-52
- [x] tested in stage with https://lms.stage.nelp.gov.sa/ and  https://edunext.stage.nelp.gov.sa/

### DIFF
--- a/lms/djangoapps/course_home_api/toggles.py
+++ b/lms/djangoapps/course_home_api/toggles.py
@@ -4,6 +4,7 @@ Toggles for course home experience.
 
 from edx_toggles.toggles import LegacyWaffleFlagNamespace
 
+from lms.lib.utils import use_learning_legacy_frontend
 from openedx.core.djangoapps.waffle_utils import CourseWaffleFlag
 
 WAFFLE_FLAG_NAMESPACE = LegacyWaffleFlagNamespace(name='course_home')
@@ -26,7 +27,10 @@ COURSE_HOME_USE_LEGACY_FRONTEND = CourseWaffleFlag(WAFFLE_FLAG_NAMESPACE, 'cours
 
 
 def course_home_legacy_is_active(course_key):
-    return COURSE_HOME_USE_LEGACY_FRONTEND.is_enabled(course_key) or course_key.deprecated
+    return (
+        use_learning_legacy_frontend() and
+        (COURSE_HOME_USE_LEGACY_FRONTEND.is_enabled(course_key) or course_key.deprecated)
+    )
 
 
 def course_home_mfe_progress_tab_is_active(course_key):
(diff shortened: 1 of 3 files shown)


lms/djangoapps/course_home_api/toggles.py
```diff
@@ -26,7 +26,11 @@
 
 
 def course_home_legacy_is_active(course_key):
-    return COURSE_HOME_USE_LEGACY_FRONTEND.is_enabled(course_key) or course_key.deprecated
+    from lms.lib.utils import use_learning_legacy_frontend
```
andrey-canon: why here instead of the top of the file ?

lms/lib/utils.py
```diff
@@ -44,3 +44,16 @@ def is_unit(xblock):
...
+def use_learning_legacy_frontend() -> bool:
+    """
+    Checks in django settings if use the learning legacy
+    frontend instead the learning mfe.
+
+    Returns:
+        True if the legacy frontend setting is activated, by default
+        legacy would be activated.
+    """
+    from django.conf import settings
```
andrey-canon: same comment  as above

lms/djangoapps/courseware/toggles.py
```diff
@@ -127,7 +127,8 @@ def courseware_mfe_is_active(course_key: CourseKey) -> bool:
         return False
     # NO: MFE courseware can be disabled for users/courses/globally via this
     #     Waffle flag.
-    if COURSEWARE_USE_LEGACY_FRONTEND.is_enabled(course_key):
+    from lms.lib.utils import use_learning_legacy_frontend
```
andrey-canon: same comment as above

lms/lib/utils.py
johanseto: here is because I want to load the setting each time the function is called. And also avoid circular imports, may be wasn't presented but was for in case. https://github.com/eduNEXT/edunext-platform/pull/733/files#diff-69936e1a963cf5b81fd008890439a3ece0d241a488fe56e45df70cac5e0b55c0L33

andrey-canon (approved):

johanseto:
I have updated the imports and seem working. 
Peek 2023-02-01 10-03
 So I gonna proceed to merge after checks.

PR #771, opened 2023-07-23
Binary gender in Saudi Arabia

shadinaif:
## Description

According to Saudi Arabia's culture and laws, we cannot have a third option in the gender field `Other/Prefer Not to Say`. We're removing it from their branch

This change shouldn't have any effect on the platform's behavior

## Supporting information

Check Slack

## Testing instructions

None

## Deadline

very urgent

## Other information

None


shadinaif:
@OmarIthawi please review

omar-nelc (approved):
Thanks Shadi for addressing this.

johanseto:
@shadinaif thanks for you recomendations. 
We apply the same approach you propose but in other parts.
So for this case in order to don't affect the platform and for example migrations checks, we have applied the changes to eox-nelp plugin.
https://github.com/eduNEXT/eox-nelp/pull/71
So there also you can check the exact change you proposed https://github.com/eduNEXT/eox-nelp/pull/71/files#diff-cb3df7bae1d68b7cad674a0068109f93517451cbd24c54a7f9928e6b4804fa33R29
without the migrations.

shadinaif:
This is perfect @johanv26 , thank you!

closing this one

PR #774, opened 2023-08-03
Merge pull request #32071 from openedx/jhynes/APER-2344

andrey-canon:
## Description

This is a simple backport of the models files from  https://github.com/openedx/edx-platform/pull/32071


johanseto (commented):
The toggles and flags doesnt need to be backported. Why?

andrey-canon:
This change is not compatible  with the current openedx-events version therefore a different approach will be applied

PR #777, opened 2023-09-06
feat: add nelp strain dispatcher

johanseto:
# Description 
with this configuration, we only have to configure the PAT for the action.

## Other information

This PAT has to be configured with someone that has permission of this.


johanseto:
@andrey-canon Token was already configured.
image

andrey-canon (approved):

PR #779, opened 2023-09-19
feat: this allows to change the editable behavior for the full name f…

andrey-canon:
## Description

this allows to change the editable behavior for the full name field

https://edunext.atlassian.net/browse/FUTUREX-521

## Testing instructions

Go to the account settings and check that the full name  can not be edited

### DIFF
--- a/lms/static/js/student_account/views/account_settings_factory.js
+++ b/lms/static/js/student_account/views/account_settings_factory.js
@@ -31,7 +31,8 @@
             extendedProfileFields,
             displayAccountDeletion,
             isSecondaryEmailFeatureEnabled,
-            betaLanguage
+            betaLanguage,
+            allowFullNameChange
         ) {
             var $accountSettingsElement, userAccountModel, userPreferencesModel, aboutSectionsData,
                 accountsSectionData, ordersSectionData, accountSettingsView, showAccountSettingsPage,
@@ -103,7 +104,7 @@
                 helpMessage: gettext('The name that is used for ID verification and that appears on your certificates.'),  // eslint-disable-line max-len,
                 persistChanges: true
             };
-            if (syncLearnerProfileData && enterpriseReadonlyAccountFields.fields.indexOf('name') !== -1) {
+            if (!allowFullNameChange || (syncLearnerProfileData && enterpriseReadonlyAccountFields.fields.indexOf('name') !== -1)) {
                 fullnameFieldView = {
                     view: new AccountSettingsFieldViews.ReadonlyFieldView(fullNameFieldData)
                 };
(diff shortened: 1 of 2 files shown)


johanseto (commented):
I am testing in stage, but I am trying to make it editable via tenant config. But isn't working
image

andrey-canon:
> I am testing in stage, but I am trying to make it editable via tenant config. But isn't working 

Because you are using dga  so you should use this branch https://github.com/nelc/nelp-edx-theme/tree/and/add_full_name_change or remove that file from the nelp-edx-theme, that is the change that I will include in another pr since that file is an exact copy of the lms version, another option is to test with the bragi theme

johanseto:
Sure that have sense, but In bragi should work ?
Screenshot from 2023-09-20 12-03-28

johanseto:
Actually is only seems working for vainilla theme.
image

johanseto:
But I dont understand why...?
image

johanseto:
> But I dont understand why...? image

Maybe it could be the other part of the if 🤔

andrey-canon:
> > But I dont understand why...? 
> 
> Maybe it could be the other part of the if 🤔

I guess bragi is using the dga template instead of the lms template

johanseto:
> > > But I dont understand why...?
> > 
> > 
> > Maybe it could be the other part of the if 🤔
> 
> I guess bragi is using the dga template instead of the lms template

Wtf so we also need a PR for bragi

andrey-canon:
> > > > But I dont understand why...?
> > > 
> > > 
> > > Maybe it could be the other part of the if 🤔
> > 
> > 
> > I guess bragi is using the dga template instead of the lms template
> 
> Wtf so we also need a PR for bragi

nop, I just set `"EOX_THEMING_DEFAULT_THEME_NAME": "",`

johanseto:
> > > > > But I dont understand why...?
> > > > 
> > > > 
> > > > Maybe it could be the other part of the if 🤔
> > > 
> > > 
> > > I guess bragi is using the dga template instead of the lms template
> > 
> > 
> > Wtf so we also need a PR for bragi
> 
> nop, I just set `"EOX_THEMING_DEFAULT_THEME_NAME": "",`

I am having problem with that change in the lms,
Actually that change in `""` could cause too many problems and could break the lms pod.
Screenshot from 2023-09-20 13-15-29
Also that error seems  created a lot of lms. 
image

Changing it to `bragi` work at supposed and also don't break.

Peek 2023-09-20 13-18

johanseto (approved):

andrey-canon:
> ause too many problems and could break the lms pod.

However when I changed that to bragi that didn't work :(, any way let's keep the full name not editable by default, then we can update the nelp-edx-theme to fully support this

PR #790, opened 2023-12-07
fix: first language visit

johanseto:
## Description

This fix is based in the method `get_language_cookie` default value. https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/lang_pref/helpers.py#L9

So in the first visit if the cookie doesnt exist, the language would be the configure by settings using `LANGUAGE_CODE`.If it is not defined, then would be `None` that is the default of the method.

## Supporting information

Jira ticket:
https://edunext.atlassian.net/jira/software/c/projects/FUTUREX/boards/36?selectedIssue=FUTUREX-640

## Testing instructions

Please provide detailed step-by-step instructions for testing this change.

### DIFF
--- a/openedx/core/djangoapps/lang_pref/middleware.py
+++ b/openedx/core/djangoapps/lang_pref/middleware.py
@@ -2,7 +2,7 @@
 Middleware for Language Preferences
 """
 
-
+from django.conf import settings
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.translation import LANGUAGE_SESSION_KEY
 from django.utils.translation.trans_real import parse_accept_lang_header
@@ -29,7 +29,7 @@ class LanguagePreferenceMiddleware(MiddlewareMixin):
         If a user's UserPreference contains a language preference, use the user's preference.
         Save the current language preference cookie as the user's preferred language.
         """
-        cookie_lang = lang_pref_helpers.get_language_cookie(request)
+        cookie_lang = lang_pref_helpers.get_language_cookie(request, getattr(settings, "LANGUAGE_CODE", None))
         if cookie_lang:
             if request.user.is_authenticated:
                 # DarkLangMiddleware will take care of this so don't change anything


johanseto:
@andrey-canon I was trying to replicate this error in my local env and stage and I didn't reached out. 
So I had to read the other [middleware](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/dark_lang/middleware.py#L92) of dark_lang and I found that this language issues could be available if the [clean-headers](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/dark_lang/middleware.py#L123-L141) clean the header "HTTP_ACCEPT_LANGUAGE". If the model of dark lhas is configurable empty in `released languages` that header become `''`. 
new_accept_header


image




And if that header is 'empty'  then Django  in the locale middleware uses the setting   `LANGUAGE_CODE` value.
https://github.com/django/django/blob/main/django/middleware/locale.py#L34

## Before
[Peek 2023-12-07 14-50](url)
## After
Peek 2023-12-07 14-54

openedx/core/djangoapps/lang_pref/middleware.py
```diff
@@ -29,7 +29,7 @@ def process_request(self, request):
         If a user's UserPreference contains a language preference, use the user's preference.
         Save the current language preference cookie as the user's preferred language.
         """
-        cookie_lang = lang_pref_helpers.get_language_cookie(request)
+        cookie_lang = lang_pref_helpers.get_language_cookie(request, getattr(settings, "LANGUAGE_CODE", None))
```
OmarIthawi, line 32: Thanks @johanv26. I believe you've spotted the right place.

> If the model of dark lhas is configurable empty in released languages that header become ''.

I'm not sure if this is an option because DGA requires two languages with Arabic by default. We also might have sites that are English by default. What do you think?

> So I had to read the other [middleware](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/dark_lang/middleware.py#L92) of dark_lang and I found that this language issues could be available if the [clean-headers](https://github.com/openedx/edx-platform/blob/master/openedx/core/djangoapps/dark_lang/middleware.py#L123-L141) clean the header "HTTP_ACCEPT_LANGUAGE".

I believe this is a good solution. Replacing `HTTP_ACCEPT_LANGUAGE` with `settings.LANGUAGE_CODE` is what we need, but only for non-API URLs because mobile would be Arabic-only otherwise.

I think `lang_pref_helpers.get_language_cookie(request, getattr(settings, "LANGUAGE_CODE", None))` is better than dropping `HTTP_ACCEPT_LANGUAGE` because it affects only web.

As an optional refactoring note: My understanding that this is either should be an Open edX plugin by upgrading[ LocalizerX (Hawthorn)](https://github.com/appsembler/LocalizerX/) or putting it inside https://github.com/eduNEXT/eox-nelp but not in the `eduNEXT/edunext-platform`. But I'll leave that decision up to you.

PR #809, opened 2024-02-27
feat: create a new setting USE_DISCUSSIONS_MFE

luisfelipec95:
## Description

This PR adds a new setting `USE_DISCUSSIONS_MFE` that will allow Discussions MFE to be enabled or disabled in the tenant

## Supporting information

[JIRA Issue DS-806](https://edunext.atlassian.net/browse/DS-806)

## Testing instructions

1. Run a Palm environment using [TVM](https://tvm.docs.edunext.co/en/latest/tvm_quickstart.html#step-by-step) or the [Tutor docs](https://docs.tutor.edly.io/install.html) in its [Palm version](https://pypi.org/project/tutor/16.1.8/)
2. Use this branch as your edx-platform version adding in the config.yml the following
  ``` yml
  EDX_PLATFORM_REPOSITORY: https://github.com/eduNEXT/edunext-platform.git
  EDX_PLATFORM_VERSION: lfc/enable-discussions-mfe
  ```
3. Install tutor-mfe directly in the Tutor environment `tutor plugins install mfe`, and enable it `tutor plugins enable mfe` 
4. `tutor plugins install forum`
5. Update the environment settings `tutor config save`
6. `tutor dev launch`
7. Go to both container `tutor dev exec lms/cms bash`, install eox-tenant `pip install eox-tenant`, and run the migrations `./manage.py lms makemigrations` and `./manage.py lms migrate`
8. Restart the containers `tutor dev restart`
9. Go to the admin panel and create 2 new routes:
  - **Domain name:** site1.local.overhang.io
    ``` json
    {
        "DISCUSSIONS_MICROFRONTEND_URL": "http://site1.local.overhang.io:2002/discussions",
        "EDNX_USE_SIGNAL": true,
        "LEARNING_MICROFRONTEND_URL": "http://site1.local.overhang.io:2000/learning",
        "LMS_BASE": "site1.local.overhang.io:8000",
        "LMS_ROOT_URL": "http://site1.local.overhang.io:8000",
        "MFE_CONFIG": {
            "EXAMS_BASE_URL": "",
            "LMS_BASE_URL": "http://site1.local.overhang.io",
            "LOGIN_URL": "http://site1.local.overhang.io/login",
            "LOGOUT_URL": "http://site1.local.overhang.io/logout"
        },
        "PLATFORM_NAME": "site1",
        "SITE_NAME": "site1.local.overhang.io",
        "USE_DISCUSSIONS_MFE": false,
        "course_org_filter": [
            "site1"
        ]
    }
[...]

PR #828, opened 2024-08-07
feat: create a new setting USE_DISCUSSIONS_MFE (#809)

dcoa:
This is a backport of the Flag for enable/disable discussions MFE by tenant


openedx/core/djangoapps/discussions/utils.py
```diff
@@ -191,3 +193,22 @@ def get_course_division_scheme(course_discussion_settings: CourseDiscussionSetti
...
+
+def use_discussions_mfe(org) -> bool:
+    """
+    Checks with the org if the tenant enables the
+    Discussions MFE.
+    Returns:
+        True if the MFE setting is activated, by default
+        the MFE is deactivated
+    """
+    ENABLE_MFE_FOR_TESTING = SettingDictToggle(
```
magajh, line 206: Maybe I'm missing some context here, but is there a particular reason why this setting is named `ENABLE_MFE_FOR_TESTING`? The name doesn't specify that it's for the Discussions MFE, and including the word 'testing' also seems out of place

mariajgrimaldi:
Is there a reason we'd want to disable the discussions MFE by tenant? Can we explain it in the cover letter? Thanks!

dcoa:
I will close this understanding we decided to use edx-platform as upstream is.

